PR #17, opened 2013-06-06
[BC Break] Fix doctrine iterators for sf2.2 & 2.3 compatibilty

lerouquin:
Replaces the deprecated `Symfony\Component\Form\Util\PropertyPath` using
`Symfony\Component\PropertyAccess\PropertyAccess` new method of
accessing properties, making it compatible with sf2.3 (and thus 2.2)

Note that the `PropertyAccess::getPropertyAccessor();` will be deprecated in sf3.0 in favor of `createPropertyAccessor()` but it is kept in order to maintain the 2.2 compatibility.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require-dev": {
         "php": ">=5.3.0",
         "symfony/routing": "*",
-        "symfony/property-access": "*"
+        "symfony/property-access": ">=2.2.*, <3.0.0-dev"
     },
     "suggest": {
         "ext-curl": "*",


Hoannv:
@rande SF just updated SF2.3. PropertyPath was moved from https://github.com/symfony/symfony/tree/2.3/src/Symfony/Component/Form/Util to https://github.com/symfony/symfony/tree/2.3/src/Symfony/Component/PropertyAccess

Class 'Symfony\Component\Form\Util\PropertyPath' not found in sonata-project/exporter/lib/Exporter/Source/DoctrineORMQuerySourceIterator.php on line 47

PR #50, opened 2015-08-12
Replace boolean's value by yes/no

chalasr:
I am working on a sonata project and I had to export datas.
All yes/no questions in my forms are stored as boolean, but on export sonata is returning boolean.
Maybe exist more simple alternative in front of Sonata Admin ?

### DIFF
--- a/lib/Exporter/Source/DoctrineORMQuerySourceIterator.php
+++ b/lib/Exporter/Source/DoctrineORMQuerySourceIterator.php
@@ -103,6 +103,8 @@ class DoctrineORMQuerySourceIterator implements SourceIteratorInterface
             $value = $value->format($this->dateTimeFormat);
         } elseif (is_object($value)) {
             $value = (string) $value;
+        } elseif (is_bool($value)) {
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
It's totally normal to have booleans on exported data.

BTW, your PR will be completely BC break.

At least, if you really want it, you could propose it as an option.

OskarStark:
i like this PR, we should merge this in a new major IMO

ping @rande

chalasr:
Maybe it's a better idea to propose it as an exporter optional config for admin class,

chalasr:
I will propose it as an exporter optional configuration for admin class, as stated by @Soullivaneuh

fre5h:
Don't forget to add translator inside DoctrineORMQuerySourceIterator to allow automatically translation of `yes` and `no` values.

OskarStark:
i think this should be done before, shouldn't it?

fre5h:
What exactly? In this PR boolean values come from entities and are change to yes/no values inside DoctrineORMQuerySourceIterator.

PR #55, opened 2015-11-17
Symfony 3.0 support

Ener-Getick:
This PR adds symfony 3.0 support.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -5,7 +5,6 @@ $finder = Symfony\CS\Finder\DefaultFinder::create()
 ;
 
 return Symfony\CS\Config\Config::create()
-    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
     ->fixers(array(
         '-unalign_double_arrow',
         '-unalign_equals',
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,21 +33,19 @@ matrix:
     - php: 5.6
       env: SYMFONY_VERSION=2.7.*
     - php: 5.6
-      env: SYMFONY_VERSION=2.8.*@dev
+      env: SYMFONY_VERSION=2.8.*@dev SYMFONY_DEPRECATIONS_HELPER=strict
     - php: 5.6
-      env: SYMFONY_VERSION="3.0.x-dev as 2.8"
+      env: SYMFONY_VERSION=3.0.*@dev DEPENDENCIES=dev COMPOSER_FLAGS="--prefer-stable" SYMFONY_DEPRECATIONS_HELPER=strict
   allow_failures:
     - php: 7.0
     - php: hhvm
-    - env: SYMFONY_VERSION=2.8.*@dev
-    - env: SYMFONY_VERSION="3.0.x-dev as 2.8"
 
 before_script:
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -12,20 +12,26 @@
             "homepage": "https://sonata-project.org/"
         }
     ],
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/Ener-Getick/PHP-CS-Fixer"
+        }
+    ],
```
Ener-Getick: Must be removed once https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1510 is merged.

OskarStark:
Thank you @Ener-Getick 

can you please check, why travis failed?

ping @Soullivaneuh

Ener-Getick:
I think that this is related to the constraint of ``fabpot/php-cs-fixer`` and that it will be fixed when I will revert it.

rande:
It will be better to not have this dependency inside composer.json file ... @Soullivaneuh will be happy ;)

.php_cs
```diff
@@ -5,7 +5,6 @@ $finder = Symfony\CS\Finder\DefaultFinder::create()
 ;
 
 return Symfony\CS\Config\Config::create()
-    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
```
Soullivaneuh: Why this?

.travis.yml
```diff
@@ -35,19 +35,18 @@ matrix:
     - php: 5.6
       env: SYMFONY_VERSION=2.8.*@dev
     - php: 5.6
-      env: SYMFONY_VERSION="3.0.x-dev as 2.8"
+      env: SYMFONY_VERSION="3.0.x-dev" DEPENDENCIES="dev" COMPOSER_FLAGS="--prefer-stable"
```
Soullivaneuh: Just use `3.0.*@dev`. The dev tag could be removed when SF 3.0 will be stable.

Soullivaneuh:
> It will be better to not have this dependency inside composer.json file ... @Soullivaneuh will be happy ;)

Are you talking about VCS? Yes, but @Ener-Getick plan to do it. Will merge this PR after that.

.php_cs
Ener-Getick: Because this function doesn't exist in Php-cs-fixer 0.* which is needed to test sf 2.3.

Ener-Getick: And because this is the default value.

Soullivaneuh: Yeah ok. This is why I want to completely remove this dependencies and use StyleCI.

See here: https://github.com/sonata-project/SonataAdminBundle/pull/3273

Ener-Getick: Another solution would be to use the same solution used in Php-cs-fixer (a regex replacement).

.travis.yml
```diff
@@ -27,27 +27,21 @@ matrix:
     - php: 5.3
       env: COMPOSER_FLAGS="--prefer-lowest"
     - php: 5.6
-      env: SYMFONY_VERSION=2.3.*
```
Ener-Getick: I removed this test as symfony 2.3 is already tested in the first one.

Ener-Getick:
Php-cs-fixer now supports symfony 3.0. 
This PR is ready to be merged if you have nothing else to change.

rande:
@Ener-Getick Thanks

OskarStark:
:+1:

PR #60, opened 2016-03-27
Add HeaderCommentFixer in .php_cs

chalasr:
'header_comment' rule added and all files that doesn't contain the header have been fixed.


OskarStark:
Thank you @chalasr

PR #76, opened 2016-05-17
add KnpSnappyWriter

### LINKED ISSUE #42: PDF Export Feature
I have seen an example how to export PDF by injecting the `KnpSnappyBundle`:
http://cristian-radulescu.ro/article/add-pdf-export-functionality-in-sonataadminbundle.html

But the tutorial also states that the Sonata way would be to use a `PdfWriter` class.
Are there any plans for such a service?

chalasr:
Closes #42 

### Changelog

- Add `KnpSnappyWriter` that introduces PDF export format.

### Subject

Provides PDF export feature.

### To do

- [x] KnpSnappyWriter creation
- [x] Update the tests
- [x] Add an upgrade note

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,13 +21,16 @@
         "symfony/phpunit-bridge": "^2.7 || ^3.0",
         "propel/propel1": "^1.6",
         "fabpot/php-cs-fixer": "~0.1 || ^1.0",
-        "sllh/php-cs-fixer-styleci-bridge": "^2.0"
+        "sllh/php-cs-fixer-styleci-bridge": "^2.0",
+        "knplabs/knp-snappy": "^0.4",
+        "h4cc/wkhtmltopdf-amd64": "^0.12"
     },
     "suggest": {
         "ext-curl": "*",
         "symfony/routing": "^2.3 || ^3.0",
         "symfony/property-access": "^2.3 || ^3.0",
-        "propel/propel1": "^1.6"
+        "propel/propel1": "^1.6",
+        "knplabs/knp-snappy": "For using the KnpSnappyWriter to export data in PDF format"
     },
     "autoload": {
         "psr-0": {
--- a/lib/Exporter/Writer/KnpSnappyWriter.php
+++ b/lib/Exporter/Writer/KnpSnappyWriter.php
@@ -1,0 +1,229 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Exporter\Writer;
+
+use Knp\Snappy\Pdf;
+
+/**
+ * Export data in PDF format.
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+final class KnpSnappyWriter implements WriterInterface
+{
+    /**
+     * @var string
+     */
+    private $filename;
+
+    /**
+     * @var string
+     */
+    private $wkhtmltopdf;
+
+    /**
+     * @var bool
+     */
+    private $showHeaders;
+
+    /**
+     * @var bool
+     */
+    private $showBorders;
+
+    /**
+     * @var array
+     *
+     * @see http://wkhtmltopdf.org/usage/wkhtmltopdf.txt
+     */
+    private $snappyOptions;
+
+    /**
+     * @var resource
+     */
+    private $file;
+
+    /**
+     * @var string
+     */
+    private $html = '
+<html>
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <meta name=ProgId content=Pdf.Document>
+    <meta name=Generator content="https://github.com/sonata-project/exporter">
+    <style></style>
+  </head>
+  <body>
+    [header-html]
+    <table>[content]</table>
+    [footer-html]
+  </body>
+</html>';
+
+    /**
+     * @throws \RuntimeException
+     *
+     * @param string $filename
+     * @param bool   $showHeaders
+     * @param bool   $showBorders
+     * @param array  $snappyOptions KnpSnappy options (wkhtmltopdf)
+     * @param string $wkhtmltopdf   Path of the wkhtmltopdf binary
+     */
+    public function __construct(
+        $filename,
+        $wkhtmltopdf,
+        $showHeaders = true,
+        $showBorders = true,
+        array $snappyOptions = array()
+    ) {
+        if (!is_bool($showHeaders)) {
+            throw new \InvalidArgumentException(sprintf('The second argument of "%s::__construct()" must be of type boolean, %s given', get_class($this), gettype($showHeaders)));
+        }
+
+        if (!is_bool($showBorders)) {
+            throw new \InvalidArgumentException(sprintf('The third argument of "%s::__construct()" must be of type boolean, %s given', get_class($this), gettype($showBorders)));
+        }
+
+        if (is_file($filename)) {
+            throw new \RuntimeException(sprintf('The file "%s" already exists', $filename));
+        }
+
+        if (!is_executable($wkhtmltopdf)) {
+            throw new \RuntimeException('The file "%s" doesn\'t exist or is not executable', $wkhtmltopdf);
+        }
+
+        $this->filename = $filename;
+        $this->showHeaders = $showHeaders;
+        $this->showBorders = $showBorders;
+        $this->snappyOptions = $snappyOptions;
+        $this->wkhtmltopdf = $wkhtmltopdf;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function open()
+    {
+        $this->file = fopen($this->filename, 'w', false);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function write(array $data)
+    {
+        $tmpFile = tmpfile();
+
+        if ($this->showBorders) {
+            $this->addBorders();
+        }
+
+        $this->addMarkupFromOptions();
+
+        if ($this->showHeaders) {
+            fwrite($tmpFile, $this->getHeaders(array_keys($data)));
+        }
+
+        fwrite($tmpFile, '<tr>');
+
+        foreach ($data as $value) {
+            fwrite($tmpFile, sprintf('<td>%s</td>', $value));
+        }
+
+        fwrite($tmpFile, '</tr>');
+
+        $tmpFileMetaData = stream_get_meta_data($tmpFile);
+        $tmpFilePath = $tmpFileMetaData['uri'];
+
+        fwrite($this->file, $this->generatePdf($tmpFilePath));
+
+        fclose($tmpFile);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function close()
+    {
+        fclose($this->file);
+    }
+
+    /**
+     * Generates a PDF using KnpSnappy.
+     *
+     * @param string $htmlFilename.
+     *
+     * @return string The generated PDF content.
+     */
+    private function generatePdf($htmlFilename)
+    {
+        $snappy = new Pdf();
+        $snappy->setBinary($this->wkhtmltopdf);
+
+        foreach ($this->snappyOptions as $key => $value) {
+            $snappy->setOption($key, $value);
+        }
+
+        $this->html = str_replace('[content]', file_get_contents($htmlFilename), $this->html);
+
+        return $snappy->getOutputFromHtml($this->html);
+    }
+
+    /**
+     * Add style for bordered table.
+     */
+    private function addBorders()
+    {
+        $borderedStyle = '<style>table{border-collapse:collapse;width:90%;}th,td{border:1px solid black;}</style>';
+
+        $this->html = str_replace('<style></style>', $borderedStyle, $this->html);
+    }
+
+    /**
+     * @param array $data
+     *
+     * @return string
+     */
+    private function getHeaders(array $columnNames)
+    {
+        $row = '';
+
+        foreach ($columnNames as $name) {
+            $row .= sprintf('<th>%s</th>', $name);
+        }
+
+        return sprintf('<tr>%s</tr>', $row);
+    }
+
+    /**
+     * Handles header-html/footer-html wkhtmltopdf options.
+     */
+    private function addMarkupFromOptions()
+    {
+        $markupOptions = array('header-html', 'footer-html');
+
+        foreach ($markupOptions as $key) {
+            if (!isset($this->snappyOptions[$key])) {
+                $this->snappyOptions[$key] = '';
+            }
+
+            $this->html = str_replace(
+                sprintf('[%s]', $key),
+                $this->snappyOptions[$key],
+                $this->html
+            );
+
+            unset($this->snappyOptions[$key]);
+        }
+    }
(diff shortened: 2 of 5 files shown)


lib/Exporter/Writer/KnpSnappyWriter.php
```diff
@@ -0,0 +1,216 @@
...
+     * {@inheritdoc}
+     */
+    public function close()
+    {
+        fclose($this->file);
+    }
+
+    /**
+     * Generates a Pdf using KnpSnappy.
+     */
```
OskarStark: `@param` and `@return` missing

lib/Exporter/Writer/KnpSnappyWriter.php
```diff
@@ -0,0 +1,216 @@
...
+
+        return $snappy->getOutputFromHtml(str_replace('[content]', file_get_contents($tmpFilePath), self::$template));
+    }
+
+    /**
+     * @param $data
+     *
+     * @return array mixed
+     */
+    protected function getHeaders($data)
```
OskarStark: can we typehint this to `array` ?

lib/Exporter/Writer/KnpSnappyWriter.php
```diff
@@ -0,0 +1,216 @@
...
+
+        foreach ($data as $header => $value) {
+            $row .= sprintf('<th>%s</th>', $header);
+        }
+
+        return sprintf('<tr>%s</tr>', $row);
+    }
+
+    /**
+     */
```
OskarStark: phpdoc wrong

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
```diff
@@ -0,0 +1,81 @@
...
+     * @var string
+     */
+    protected $wkhtmltopdfBinary;
+
+    public function setUp()
+    {
+        $this->filename = 'foobar.pdf';
+        $this->wkhtmltopdfBinary = __DIR__.'/../../../../vendor/h4cc/wkhtmltopdf-amd64/bin/wkhtmltopdf-amd64';
+
+        if (!is_file($this->wkhtmltopdfBinary)) {
```
OskarStark: should be executable, too imo

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
```diff
@@ -0,0 +1,81 @@
...
+    protected $filename;
+
+    /**
+     * @var string
+     */
+    protected $wkhtmltopdfBinary;
+
+    public function setUp()
+    {
+        $this->filename = 'foobar.pdf';
```
OskarStark: can we do this with a virtual filesystem?

chalasr:
@OskarStark I can't figure out the StyleCI warnings, can you help me?

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
chalasr: I tried to implement it in the SonataAdminBundle Exporter (`php://output`), works well.

lib/Exporter/Writer/KnpSnappyWriter.php
```diff
@@ -0,0 +1,216 @@
...
+     * @param bool   $showBorders
+     * @param array  $snappyOptions     KnpSnappy options (wkhtmltopdf)
+     * @param string $wkhtmltopdfBinary Path of the wkhtmltopdf binary
+     */
+    public function __construct(
+        $filename,
+        $wkhtmltopdfBinary,
+        $showHeaders = true,
+        $showBorders = true,
+        array $snappyOptions = null
```
OskarStark: wouldn't it be easier to set this do `array()` as default?

composer.json
```diff
@@ -21,7 +21,8 @@
         "symfony/phpunit-bridge": "^2.7 || ^3.0",
         "propel/propel1": "^1.6",
         "fabpot/php-cs-fixer": "~0.1 || ^1.0",
-        "sllh/php-cs-fixer-styleci-bridge": "^2.0"
+        "sllh/php-cs-fixer-styleci-bridge": "^2.0",
+        "knplabs/knp-snappy": "~0.4"
```
Soullivaneuh: * `^0.4`
* Keys should be added on `suggest` and `conflict` sections.

lib/Exporter/Writer/KnpSnappyWriter.php
```diff
@@ -0,0 +1,216 @@
...
+namespace Exporter\Writer;
+
+use Knp\Snappy\Pdf;
+
+/**
+ * Export data in PDF format.
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+class KnpSnappyWriter implements WriterInterface
```
Soullivaneuh: I think this should be `final`.

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
```diff
@@ -0,0 +1,82 @@
...
+namespace Exporter\Test\Source;
+
+use Exporter\Writer\KnpSnappyWriter;
+
+/**
+ * Tests the KnpSnapyWriter Writer.
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+class KnpSnappyWriterTest extends \PHPUnit_Framework_TestCase
```
Soullivaneuh: `final`

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
chalasr: Is it what you are asking?

Soullivaneuh:
> @OskarStark I can't figure out the StyleCI warnings, can you help me?

Two solution for this:

### PHP-CS-Fixer

Run `php-cs-fixer fix -v` command and commit your change

### Git apply

Go to the StyleCI report and copy the link of "Download RAW".

Then, on command line:

```bash
wget https://styleci.io/analyses/XVB0r3/diff
git apply diff
rm diff
git commit
git push
```

chalasr:
Thank you @Soullivaneuh

composer.json
```diff
@@ -21,13 +21,17 @@
...
+        "knplabs/knp-snappy": "^0.4",
+        "h4cc/wkhtmltopdf-amd64": "^0.12"
     },
     "suggest": {
         "ext-curl": "*",
         "symfony/routing": "^2.3 || ^3.0",
         "symfony/property-access": "^2.3 || ^3.0",
-        "propel/propel1": "^1.6"
+        "propel/propel1": "^1.6",
+        "knplabs/knp-snappy": "^0.4",
```
core23: Please add a text information, not a version here.

See https://getcomposer.org/doc/04-schema.md#suggest

core23: Version information should be mention under `conflict`
https://getcomposer.org/doc/04-schema.md#conflict

chalasr: This was my first intention; but as other ones don't have any description, plus there is no real conflict between versions.

chalasr:
Doc updated, more tests added, line-notes fixed, suggestion phrase set, branch rebased.

The wkhtmltopdf has also been added as dev dependency only for running tests.
These tests make the suite duration longer (from ~15min to ~35min), if wkhtmltopdf come to be removed from requirements, tests would then be skipped.

core23:
LGTM

chalasr:
Whah are you expecting?
I updated the possible outputs in the doc.
There is no usage documented for any writer, so do it for only one wouldn't  make sense.
Am I missing something?

lib/Exporter/Writer/KnpSnappyWriter.php
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function write(array $data)
+    {
+        $tmpFile = tmpfile();
+
+        if (true === $this->showBorders) {
```
greg0ire: @Soullivaneuh : about strict checks, here is an example. Here, if `$this->showBorders === 1` instead of `true`, no borders are added. Silently. If `$this->showBorders === 0` instead of `false`, the code works, which means that using this strict checking does not even ensure the method cannot be misused with something else than a boolean

@Soullivaneuh @chalasr IMO strict checking is pointless, `if ($this->showBorders)` would be way more natural to read. If you want to do strict checking, then do it before, like this : 

```
if (!is_bool($this->showBorders)) {
// … throw exception or deprecation warning saying that we will use type hinting when dropping php < 7
}
```

lib/Exporter/Writer/KnpSnappyWriter.php
```diff
@@ -0,0 +1,221 @@
...
+     */
+    private function getHeaders(array $data)
+    {
+        $row = '';
+
+        foreach ($data as $header => $value) {
+            $row .= sprintf('<th>%s</th>', $header);
+        }
+
+        return sprintf('<tr>%s</tr>', $row);
```
greg0ire: You're not using a templating system, but you can still get syntactic coloration for html, use templating syntax and avoid concatenation by doing something like this :
```php
<?php
ob_start();?>
<tr>
   <?php foreach ($data as $header => $value) : ?>
      <tr><?= $value ;></tr>
   <?php enforeach; ?>
</tr>
return ob_get_clean();
```

lib/Exporter/Writer/KnpSnappyWriter.php
```diff
@@ -0,0 +1,221 @@
...
+
+        return $snappy->getOutputFromHtml($this->html);
+    }
+
+    /**
+     * @param array $data
+     *
+     * @return string
+     */
+    private function getHeaders(array $data)
```
greg0ire: Please do not use `$data` as a variable name. Try to be more precise.

core23:
> What are you expecting?
> I updated the possible outputs in the doc.

Sorry, I missed that. Should be okay for now, but we should add some more information in general for this bundle.

lib/Exporter/Writer/KnpSnappyWriter.php
```diff
@@ -0,0 +1,221 @@
...
+     */
+    public function open()
+    {
+        $this->file = fopen($this->filename, 'w', false);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function write(array $data)
```
greg0ire: `$data` but… not your fault, it must be part of the interface I presume T__T

lib/Exporter/Writer/KnpSnappyWriter.php
chalasr: `$data` is used by `WriterInterface::write()` and all the Writer classes (even for methods such as "addHeaders". I really don't see how I can be more precise without being wrong.

lib/Exporter/Writer/KnpSnappyWriter.php
chalasr: The code you given doesn't work as is. I prefer not having markup coloration than having several  PHP open/close tags. The code would become uglier than it already is and harder to read (my opinion).

lib/Exporter/Writer/KnpSnappyWriter.php
```diff
@@ -0,0 +1,221 @@
...
+    /**
+     * @param array $data
+     *
+     * @return string
+     */
+    private function getHeaders(array $data)
+    {
+        $row = '';
+
+        foreach ($data as $header => $value) {
```
greg0ire: Here you are not using `$value`,  so you might as well use `array_keys()`

lib/Exporter/Writer/KnpSnappyWriter.php
greg0ire: `$data` is the most imprecise variable name there is… looking at your code, I can retro-engineer that it is a collection of values which keys are headers so… it is a row? Or a line ?

lib/Exporter/Writer/KnpSnappyWriter.php
greg0ire: > The code you given doesn't work as is.

That's why I said "something like this" ;)

It is in the natural order, and what you are doing is templating for a header, so you should use something that looks like a template. The code is simple at the moment, but if it becomes more complex, concatenation and sprintf will no longer be acceptable. That's why we don't use concatenation and sprintf for templates.

chalasr:
The code here is sufficient to me (even using concatenation to build the template, as done in the XLSWriter).
I made everything to fit your contributing guidelines (even almost 1year after the war) to provide this missing feature.
So, sorry for this but I can't give more efforts for finally nothing.
Feel free to merge it, improve it or close it.

greg0ire:
Well I guess it is already quite good, so LGTM!

test/Exporter/Test/Writer/GsaFeedWriterTest.php
```diff
@@ -102,7 +102,7 @@ public function testSimpleWrite()
         $generatedFiles = $this->getFiles();
 
         $this->assertCount(1, $generatedFiles);
-        $this->assertEquals($this->folder.'/feed_00001.xml', $generatedFiles[0]);
+        // $this->assertEquals($this->folder.'/feed_00001.xml', $generatedFiles[0]);
```
greg0ire: We will have to remove this

chalasr: I reverted it, this test causes a fatal error on OSX (or at least on the two different macs I use):

> 1) Exporter\Test\Source\GsaFeedWriterTest::testSimpleWrite
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'/var/folders/86/n5nbdtfd6yg2s963qlhtnb5m0000gn/T/sonata_exporter_test/feed_00001.xml'
+'/private/var/folders/86/n5nbdtfd6yg2s963qlhtnb5m0000gn/T/sonata_exporter_test/feed_00001.xml'

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * @var string
+     */
+    private $wkhtmltopdfBinary;
+
+    public function setUp()
+    {
+        $this->filename = 'foobar.pdf';
+        // $this->wkhtmltopdfBinary = __DIR__.'/../../../../vendor/h4cc/wkhtmltopdf-amd64/bin/wkhtmltopdf-amd64';
+        $this->wkhtmltopdfBinary = '/usr/local/bin/wkhtmltopdf';
```
Soullivaneuh: Could not this be configurable? Or just set `'wkhtmltopdf'` considering the binary is under a path...

greg0ire:
If this is not merged before next week, I'll take it from here and make my own PR based on this.

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
chalasr: > Could not this be configurable?

Regarding the test only or the writer `wkhtmltopdfBinary` argument itself? 
For the test, I can set the binary path as an environment variable or something like to avoid the hard-coded path. If you have a better suggestion...

greg0ire:
> If this is not merged before next week, I'll take it from here and make my own PR based on this.

In fact I can take it on right now.

test/Exporter/Test/Writer/GsaFeedWriterTest.php
greg0ire: Thank you for that

greg0ire:
Closed in favor of #82 . Thanks a lot for your patience @chalasr !

OskarStark:
#82 is closed too, the new PR can be found here: #86

PR #86, opened 2016-05-19
Add KnpSnappyWriter

greg0ire:
### Changelog

```markdown
### Added
- Added pdf writer based on [knp/snappy](https://packagist.org/packages/knplabs/knp-snappy)
```

### Subject

This PR is my take on #76 , thanks @chalasr for what you already did

### To do

~~- [ ] consider relying on knp/snappy-bundle to inject the dependency~~ this is not a symfony bundle so we can't
- [x] fix DIP violation (dependency on Pdf is not injected)
- [x] write real unit tests (the tests that already exist are integration tests). Implies applying this : https://phpunit.de/manual/current/en/test-doubles.html#test-doubles.mocking-the-filesystem
- [x] document usage

### DIFF
--- a/test/Exporter/Test/Integration/Writer/KnpSnappyWriterTest.php
+++ b/test/Exporter/Test/Integration/Writer/KnpSnappyWriterTest.php
@@ -1,0 +1,145 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Exporter\Test\Integration\Writer;
+
+use Exporter\Writer\KnpSnappyWriter;
+use Knp\Snappy\Pdf;
+
+/**
+ * Tests the KnpSnapyWriter.
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+final class KnpSnappyWriterTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @var string relative path to the test file
+     */
+    private $filename;
+
+    /**
+     * @var string absolute path to the wkhtmltopdf binary
+     */
+    private $wkhtmltopdf;
+
+    public function setUp()
+    {
+        $this->pdfGenerator = new Pdf();
+        $this->filename = 'foobar.pdf';
+        $this->wkhtmltopdf = __DIR__.'/../../../../../vendor/h4cc/wkhtmltopdf-amd64/bin/wkhtmltopdf-amd64';
+
+        if (!is_executable($this->wkhtmltopdf)) {
+            $this->markTestSkipped('The wkhtmltopdf binary is not available');
+        }
+
+        if (is_file($this->filename)) {
+            unlink($this->filename);
+        }
+    }
+
+    public function tearDown()
+    {
+        unlink($this->filename);
+    }
+
+    public function testWriteWithoutArgs()
+    {
+        $writer = new KnpSnappyWriter($this->pdfGenerator, $this->filename, $this->wkhtmltopdf, false);
+        $writer->open();
+        $writer->write(array('john "2', 'doe', '1'));
+        $writer->close();
+
+        $this->assertFileExists($this->filename);
+    }
+
+    public function testWithHeaders()
+    {
+        $writer = new KnpSnappyWriter($this->pdfGenerator, $this->filename, $this->wkhtmltopdf, true, false);
+
+        $writer->open();
+        $writer->write(array('firstname' => 'john "2', 'surname' => 'doe', 'year' => '1'));
+        $writer->close();
+
+        $this->assertAttributeContains('firstname', 'html', $writer);
+        $this->assertFileExists($this->filename);
+    }
+
+    public function testWithBorders()
+    {
+        $borderedStyle = '<style>table{border-collapse:collapse;width:90%;}th,td{border:1px solid black;}</style>';
+        $writer = new KnpSnappyWriter($this->pdfGenerator, $this->filename, $this->wkhtmltopdf, false, true);
+
+        $writer->open();
+        $writer->write(array('john "2', 'doe', '1'));
+        $writer->close();
+
+        $this->assertAttributeContains($borderedStyle, 'html', $writer);
+        $this->assertFileExists($this->filename);
+    }
+
+    public function testWithoutBorders()
+    {
+        $borderedStyle = '<style>table{border-collapse:collapse;width:90%;}th,td{border:1px solid black;}</style>';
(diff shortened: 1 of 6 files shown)


Soullivaneuh:
`knplabs/knp-snappy` should be also added on `conflict` section.

test/Exporter/Test/Integration/Writer/KnpSnappyWriterTest.php
```diff
@@ -0,0 +1,145 @@
...
+    private $wkhtmltopdf;
+
+    public function setUp()
+    {
+        $this->pdfGenerator = new Pdf();
+        $this->filename = 'foobar.pdf';
+        $this->wkhtmltopdf = __DIR__.'/../../../../../vendor/h4cc/wkhtmltopdf-amd64/bin/wkhtmltopdf-amd64';
+
+        if (!is_executable($this->wkhtmltopdf)) {
+            $this->markTestSkipped('The wkhtmltopdf binary is not available');
```
Soullivaneuh: Should not be a failed? AFAIK, the binary is downloaded from composer so it should be always available for tests.

test/Exporter/Test/Integration/Writer/KnpSnappyWriterTest.php
```diff
@@ -0,0 +1,145 @@
...
+     * @var string relative path to the test file
+     */
+    private $filename;
+
+    /**
+     * @var string absolute path to the wkhtmltopdf binary
+     */
+    private $wkhtmltopdf;
+
+    public function setUp()
```
Soullivaneuh: To be checked but I think `setUp` and `tearDown` are protected.

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
```diff
@@ -0,0 +1,82 @@
...
+// Use same namespace as system under test to mock native functions
+
+namespace Exporter\Writer;
+
+use org\bovigo\vfs\vfsStream;
+use org\bovigo\vfs\vfsStreamDirectory;
+use org\bovigo\vfs\vfsStreamWrapper;
+use Prophecy\Argument;
+
+function is_executable()
```
Soullivaneuh: WTH is that?

Soullivaneuh:
Not sure about `Prophecy` mixing `getMock` system. Maybe we should discuss about what we should apply for the future in order to keep consistency?

test/Exporter/Test/Integration/Writer/KnpSnappyWriterTest.php
greg0ire: Wouldn't there be an exception in that case?

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
greg0ire: An old trick to mock a native php function ;) See the comment line 12

greg0ire:
> Not sure about Prophecy mixing getMock system. Maybe we should discuss about what we should apply for the future in order to keep consistency?

There is not call to `getMock` in my test, and I think we should always go with Prophecy.

test/Exporter/Test/Integration/Writer/KnpSnappyWriterTest.php
greg0ire: Hmmm… probably a legacy from before doing this with Composer ? @chalasr , can you help ?

chalasr: In fact, h4cc binaries don't work on osx at all (this is why it was previously set to /usr/local/bin), and is separated in two 32/64 packages. It was a kind of security for don't break them all.

Soullivaneuh:
> There is not call to getMock in my test, and I think we should always go with Prophecy.

* You think, not all people do. This is why I would like to discuss. :-)
* I'm not talking about your test but about tests in general.

BTW, I would like to refactor some test later,it would be great to have a plan.

test/Exporter/Test/Integration/Writer/KnpSnappyWriterTest.php
Soullivaneuh: Why?

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
Soullivaneuh: Impossible to find another way? I understand the trick but wow, such ugly things! :stuck_out_tongue:

greg0ire:
Discussion is here : https://github.com/sonata-project/dev-kit/issues/89

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
greg0ire: > Impossible to find another way?

Without touching the code? Not sure, maybe the virtual system can be used there too…

So why is it so ugly? Well probably because we should use an abstraction to do that instead of doing ourselves. I don't think symfony/filesystem does this, do you know any lib that does?

test/Exporter/Test/Integration/Writer/KnpSnappyWriterTest.php
greg0ire: For some reason, I thought opening the accessibility of a method was forbidden… turns out it is not : https://3v4l.org/4XNQB . Closing it is : https://3v4l.org/0FrGI And you are right, this method should be protected

OskarStark: > To be checked but I think setUp and tearDown are protected.

thats true

rande:
I don't think this should be part of the project. Rendering PDF is more about reporting that exporting raw data (the idea of this lib). The PDF output will never match client's requirements. 

This code should rely on a dedicated reporting bundle, with specific output handler: princexml, wkhtml, etc ...

greg0ire:
Hmm so off-topic… I did not see that because I don't know the library that well… closing then.

test/Exporter/Test/Writer/KnpSnappyWriterTest.php
jordisala1991: Why do you need to mock is_executable if you have the executable already in the vendors folder (and if not, the test is skipped)?

chalasr: @jordisala1991 He is using a [virtual filesystem] to test the Writer class properly (without testing any dependencies). Integration tests (that test the whole behavior) are located in the Integration sub-namespace. Unfortunately (as you may know) PDF export is very hard to make "generic" and so it could lead to unexpected results in an admin context, this will not be merged.

rande:
@greg0ire It is actually my bad, see my comment: https://github.com/sonata-project/exporter/issues/42

greg0ire:
Yeah just saw that indeed. We should create sonataReportingBundle ;)

chalasr:
At first a Reporter lib could be sufficient and, in a next time, integrate it in the admin-bundle through a Reporter class like done for this lib.

PR #90, opened 2016-05-25
When applicable, provide mime type

greg0ire:
### Changelog

```markdown
### Added
- `MimeTypedWriterInterface` can be implemented to indicate the suitable `Content-Type` header and format for a writer.
```

### Subject

This PR allows us to avoid [this kind of thing](https://github.com/sonata-project/SonataCoreBundle/blob/3.x/Exporter/Exporter.php#L37)

### To do

- [x] Add `getFormatName()`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,11 @@
             "Exporter\\": "src/"
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "Exporter\\Test\\": "test/"
+        }
+    },
     "extra": {
         "branch-alias": {
             "dev-master": "1.x-dev"
--- a/docs/reference/outputs.rst
+++ b/docs/reference/outputs.rst
@@ -13,4 +13,6 @@ Several output formatters are supported:
 * Excel XML
 * XLS (MS Excel)
 
-You may also create your own. To do so, simply create a class that implements the ``Exporter\Writer\WriterInterface``.
+You may also create your own. To do so, simply create a class that implements the ``Exporter\Writer\WriterInterface``,
+or better, if you know what ``Content-Type`` header should be used along with
+your output and what format it produces, ``TypedWriterInterface``.
--- a/src/Writer/CsvWriter.php
+++ b/src/Writer/CsvWriter.php
@@ -13,7 +13,7 @@ namespace Exporter\Writer;
 
 use Exporter\Exception\InvalidDataFormatException;
 
-class CsvWriter implements WriterInterface
+class CsvWriter implements TypedWriterInterface
 {
     /**
      * @var string
@@ -76,6 +76,22 @@ class CsvWriter implements WriterInterface
         if (is_file($filename)) {
             throw new \RuntimeException(sprintf('The file %s already exist', $filename));
         }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    final public function getDefaultMimeType()
+    {
+        return 'text/csv';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    final public function getFormat()
+    {
+        return 'csv';
     }
 
     /**
--- a/src/Writer/CsvWriter.php
+++ b/src/Writer/CsvWriter.php
@@ -13,6 +13,9 @@ namespace Exporter\Writer;
 
 use Exporter\Exception\InvalidDataFormatException;
 
+/**
+ * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ */
 class CsvWriter implements TypedWriterInterface
 {
     /**
--- a/src/Writer/JsonWriter.php
+++ b/src/Writer/JsonWriter.php
@@ -11,6 +11,9 @@
 
 namespace Exporter\Writer;
 
+/**
+ * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ */
 class JsonWriter implements TypedWriterInterface
 {
     /**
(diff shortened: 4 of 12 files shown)


greg0ire:
Maybe a `getFormatName` could be added too, but it would most certainly imply finding a new name for the interface…

lib/Exporter/Writer/MimeTypedWriterInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+namespace Exporter\Writer;
+
+interface MimeTypedWriterInterface extends WriterInterface
+{
+    /**
+     * @return string the mime type of the output
+     */
+    public function getMimeType();
```
Soullivaneuh: Maybe `getDefaultMimeType`?

For example, json has multiple mime types: http://stackoverflow.com/q/477816/1731473

greg0ire: `Default` would mean you can set another, which might not be true… let's keep it simple ok? `getMimeType` really means "get the mime type we feel is the most appropriate", so I'll just add that as a description for the method.

lib/Exporter/Writer/MimeTypedWriterInterface.php
```diff
@@ -0,0 +1,31 @@
...
+
+interface MimeTypedWriterInterface extends WriterInterface
+{
+    /**
+     * There can be several mime types for a given format, this method should
+     * return the most appopriate / popular one.
+     *
+     * @return string the mime type of the output
+     */
+    public function getMimeType();
```
Soullivaneuh: > Default would mean you can set another, which might not be true… let's keep it simple ok? getMimeType really means "get the mime type we feel is the most appropriate", so I'll just add that as a description for the method.

I'm sorry but this should not be changed, you may set children methods final. :stuck_out_tongue: 

As I show to you, json for example can have multiple mime type, this is why I'm suggesting `Default`.

lib/Exporter/Writer/MimeTypedWriterInterface.php
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Returns a string best describing the format of the output (the file
+     * extension is fine, for example).
+     *
+     * @return string a string without spaces, usable in a translation string
+     */
+    public function getFormat();
```
Soullivaneuh: Why not `getFileExtension`?

greg0ire: Because the extension is not necessarily the format. For instance this method should return "yaml", not "yml", it's about describing the writer.

greg0ire: Oh BTW the goal of this method is to simplify [this](https://github.com/sonata-project/SonataCoreBundle/pull/270/files#r64661663)

Soullivaneuh:
Please add class comment with `@author` tag on `TypedWriterInterface`. You can also add authors on other class with your name and the name of @rande. :wink:

greg0ire:
@rande's name will do for writers, I merely describe what they do as opposed to implementing them

rande:
I can see the benefit of those functions, however not sure this require a new interface. Why not creating a factory like: https://github.com/sonata-project/SonataCoreBundle/blob/3.x/Exporter/Exporter.php ?

greg0ire:
The benefit is better seen here : https://github.com/sonata-project/SonataCoreBundle/pull/270 .  Once this is merged, this could be even more simplified by implementing this interface and using `getMimeType` and `getFormat` instead of maintaining the information about mime type and format in the exporter class (outside the exporter library, in the core bundle).

greg0ire:
@rande : plus with a factory, you can't inject your own writers (or existing writers, configured like you want).

Soullivaneuh:
@rande What is your final decision for this?

cc @greg0ire

greg0ire:
@Soullivaneuh : code coverage decrease, yet I don't think this needs any testing… do you?

greg0ire:
As suggested by @Soullivaneuh , I'll write an abstract test class that will be extended by implementations of this interface. It will check that a string is returned. Code will be covered, and it won't feel stupid, so everyone will be happy.

greg0ire:
Blocked until #104 is merged

greg0ire:
Alternatively, you might merge this and close #104

Soullivaneuh:
LGTM for me, but I would like @rande final decision.

Soullivaneuh:
ping @rande. Can we merge this one?

greg0ire:
@Soullivaneuh : maybe get a review from other @sonata-project/contributors instead ?

composer.json
```diff
@@ -34,6 +34,11 @@
             "Exporter": "lib/"
         }
     },
+    "autoload-dev": {
+        "psr-0": {
```
core23: PSR 0 is deprecated. Please use PSR 4

greg0ire: @core23 fixed… sort of… I think I'll make a separate PR and rebase on it if this turns out to rot.

greg0ire: See #114

greg0ire:
Rebased on #114

greg0ire:
Can we merge this, someday?

Soullivaneuh:
@greg0ire Please rebase.

greg0ire:
@Soullivaneuh done

Soullivaneuh:
No answer from @rande and the PR is good IMHO. Let's merge as is.

If something can be improved, we can do it later. :+1:

PR #96, opened 2016-05-28
Fixed wrong or missing PHPDoc

core23:
### Changelog

```markdown
### Fixed
- Fixed wrong or missing PHPDoc
```

### Subject
Some PHPDoc annotations are wrong or missing.

### DIFF
--- a/lib/Exporter/Source/AbstractXmlSourceIterator.php
+++ b/lib/Exporter/Source/AbstractXmlSourceIterator.php
@@ -18,17 +18,59 @@ namespace Exporter\Source;
  */
 abstract class AbstractXmlSourceIterator implements SourceIteratorInterface
 {
+    /**
+     * @var string
+     */
     protected $filename = null;
+
+    /**
+     * @var resource
+     */
     protected $file = null;
+
+    /**
+     * @var bool|null
+     */
     protected $hasHeaders = null;
+
+    /**
+     * @var string[]
+     */
     protected $columns = array();
+
+    /**
+     * @var resource
+     */
     protected $parser = null;
 
+    /**
+     * @var int
+     */
     protected $currentRowIndex = 0;
+
+    /**
+     * @var int
+     */
     protected $currentColumnIndex = 0;
+
+    /**
+     * @var mixed
+     */
     protected $currentRow = null;
+
+    /**
+     * @var array
+     */
     protected $bufferedRow = array();
+
+    /**
+     * @var bool
+     */
     protected $currentRowEnded = false;
+
+    /**
+     * @var int
+     */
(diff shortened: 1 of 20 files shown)


lib/Exporter/Source/AbstractXmlSourceIterator.php
```diff
@@ -18,17 +18,59 @@
...
+     */
     protected $filename = null;
+
+    /**
+     * @var resource
+     */
     protected $file = null;
+
+    /**
+     * @var bool|null
```
OskarStark: if we use `|null` here we need to add this to the other properties too

Soullivaneuh: Why? Some properties should not have `null` value.

OskarStark:
LGTM 👍

lib/Exporter/Source/ArraySourceIterator.php
```diff
@@ -13,8 +13,6 @@
 
 class ArraySourceIterator extends IteratorSourceIterator
 {
-    protected $iterator;
```
Soullivaneuh: This is BC break.

And why?

lib/Exporter/Source/AbstractXmlSourceIterator.php
OskarStark: check the property above

lib/Exporter/Writer/JsonWriter.php
```diff
@@ -13,10 +13,19 @@
 
 class JsonWriter implements WriterInterface
 {
+    /**
+     * @var string
+     */
     protected $filename;
 
+    /**
+     * @var
```
Soullivaneuh: This is the mysterious `@var`? :trollface:

lib/Exporter/Source/ArraySourceIterator.php
core23: The same property exists in `IteratorSourceIterator`, so no BC break .

lib/Exporter/Writer/JsonWriter.php
core23: It's an invisibly type ;)

lib/Exporter/Source/AbstractXmlSourceIterator.php
core23: Should I remove the `null` anotations or will we do this with another PR?

OskarStark: i would add the null property in this PR

Soullivaneuh:
:+1:

PR #104, opened 2016-05-30
Declare actual namespace.

greg0ire:
### Changelog

```markdown
### Fixed

- copy/paste mistake
```

### Subject

Respect PSR-0


Soullivaneuh:
Was those classes with the errors under a stable version?

PR #114, opened 2016-06-07
Use psr-4

greg0ire:
This simplifies the directory structure.


greg0ire:
ping @sonata-project/contributors

Soullivaneuh:
For me, moving file **is** bc break, isn't it?

Soullivaneuh:
And in this case, why not a the more generic `src` folder?

greg0ire:
> For me, moving file is bc break, isn't it?

Changing anything in a library is a BC-break : for instance, if people use the source to compute an md5 that would be used as a passphrase for a secret key encrypting their secrets, changing a character would be a BC-break. :trollface: 

So now, why do you think this would not be an acceptable BC-break? Do you think people might use `require_once` ?

greg0ire:
> And in this case, why not a the more generic src folder?

This is good and I thought of it, but discarded it because it does not have to do with PSR-4, does it ?

greg0ire:
ping @Soullivaneuh

Soullivaneuh:
> Changing anything in a library is a BC-break : for instance, if people use the source to compute an md5 that would be used as a passphrase for a secret key encrypting their secrets, changing a character would be a BC-break. :trollface:

:trollface: :trollface: :trollface: 

> So now, why do you think this would not be an acceptable BC-break? Do you think people might use require_once ?


As far as this lib is compliant only with composer, I think the BC-break is acceptable.

I would like another opinions from @sonata-project/contributors for that.

> I mean `lib` was already there.

As bad code. This does not mean this should not be changed. :wink:

greg0ire:
I'll rename lib

Soullivaneuh:
> I'll rename lib

You mean `src` ?

greg0ire:
> You mean src ?

No. I mean I'll rename `lib`. To `src`. As I said, this has nothing to do with PSR-4, but it will other people making PRs rebase once instead of twice, so… a seperate commit will do.

Soullivaneuh:
@greg0ire Even if it's not BC break and change nothing on the code behavior, I would like to merge it for a new minor with an upgrade note.

What do you think?

greg0ire:
I think you're right : pedantic is for new (pedantic) things, not bugfixes.

OskarStark:
I changed the `pedantic` to `minor`

greg0ire:
Well pedantic and minor is not incompatible AFAIK

OskarStark:
> Well pedantic and minor are not incompatible AFAIK

but using both does not make sense imo?

greg0ire:
> but using both does not make sense imo?

I don't know, what do you think @Soullivaneuh ?

greg0ire:
@Soullivaneuh can you review the upgrade note ? Not 100% of what you want…

UPGRADE-1.x.md
```diff
@@ -0,0 +1,10 @@
+UPGRADE 1.x
+===========
+
+UPGRADE FROM 1.4 to 1.5
+=======================
+
+## Changes in directory structure
+
+The `lib` directory has been renamed to the more standard `src`.
+Also the directory structure of `src` and `test` no longer follows `PSR-0`, but `PSR-4`.
```
Soullivaneuh: New line before this sentence.

Maybe also add a line telling this **should** not change anything because the library is aim to be used with composer only?

greg0ire:
ping @sonata-project/contributors

UPGRADE-1.x.md
```diff
@@ -0,0 +1,13 @@
...
+===========
+
+UPGRADE FROM 1.4 to 1.5
+=======================
+
+## Changes in directory structure
+
+The `lib` directory has been renamed to the more standard `src`.
+
+Also, the directory structure of `src` and `test` no longer follows `PSR-0`,
```
Soullivaneuh: Add on the same line:

```markdown
but `PSR-4`.
```

UPGRADE-1.x.md
```diff
@@ -0,0 +1,13 @@
...
+but `PSR-4`. This should not change anything for users since this library is
```
Soullivaneuh: One line if under 120 chars.

Soullivaneuh:
Except my comments, all good to me. :+1:

UPGRADE-1.x.md
```diff
@@ -0,0 +1,12 @@
+UPGRADE 1.x
+===========
+
+UPGRADE FROM 1.4 to 1.5
```
Soullivaneuh: Could you remove this? This is added while releasing.

greg0ire: @Soullivaneuh done

PR #116, opened 2016-06-09
Fix wrong PHPDoc type on JsonWriter::$file

Soullivaneuh:
### Subject

It should be a resource, not a string.


Soullivaneuh:
Not sure this error is related: https://travis-ci.org/sonata-project/exporter/jobs/136346189#L366

PR #131, opened 2016-10-03
allow closure to reaplace getValue method on iterator

zajca:
I am targetting this branch, because this is not BC

Fixes 1
https://github.com/sonata-project/SonataAdminBundle/issues/4151
## Changelog

``` markdown
### Added
- Added optional closure option to exporter fields definition
```
## To do
- [ ] Update the tests
- [ ] Update the documentation
- [ ] Add an upgrade note
## Subject

This will allow to pass closure with export field definition and exported value can be processed as user wish.
### Usage in sonata admin

``` php
    public function getExportFields()
    {
        $translator = $this->translator;
        return [
            'id',
            'status' => [
                'property' => 'status',
                'closure' => function ($value) use ($translator) {
                    return $translator->trans($value);
                },
            ],
   }
```

### DIFF
--- a/src/Source/DoctrineORMQuerySourceIterator.php
+++ b/src/Source/DoctrineORMQuerySourceIterator.php
@@ -64,7 +64,15 @@ class DoctrineORMQuerySourceIterator implements SourceIteratorInterface
             if (is_string($name) && is_string($field)) {
                 $this->propertyPaths[$name] = new PropertyPath($field);
             } else {
-                $this->propertyPaths[$field] = new PropertyPath($field);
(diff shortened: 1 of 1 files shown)


zajca:
Sorry, I forget about this. 
But later I solved this with custom iterator, it's bit hackish with sonata to use custom SourceIterator since whole method ```getDataSourceIterator``` must be replaced but it works.

Feel free to close this PR. I think custom source iterator is better way and base iterators should be simple.

greg0ire:
Thanks for the feedback!

PR #147, opened 2017-03-15
Improve DateTime formatting/granularity

gnat42:
I am targeting this branch, because this is a BC safe change.

## Changelog
Added ability to format a DateTime per source property

```markdown
### Added
- Added ability to format dates on a per property basis for DoctrineORM/ODM Query Sources.
```
## To do
Add tests if the basic idea is acceptable
Expand to cover any other Source interfaces that could benefit from a similar ability.

- [ ] Update the tests
- [ ] Update the documentation
- [ ] Add an upgrade note

## Subject
This adds the ability to two DoctrineQuerySourceInterators to format
each DateTime field individually. It also allows for any source interface to be more extensible.

### DIFF
--- a/src/Formatter/ArrayFormatter.php
+++ b/src/Formatter/ArrayFormatter.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Exporter\Formatter;
+
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+final class ArrayFormatter implements DataFormatterInterface
+{
+    /**
+     * @inheritDoc
+     */
+    public function supports($data)
+    {
+        return (is_array($data) || $data instanceof \Traversable);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function format($data, PropertyPath $propertyPath)
+    {
+        return null;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getPriority()
+    {
+        return 10;
+    }
+}
--- a/src/Formatter/DataFormatterInterface.php
+++ b/src/Formatter/DataFormatterInterface.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Exporter\Formatter;
+
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+/**
+ * Interface to implement for data formatters that will handle some
+ * input and format the output as desired for exporting. For example
+ * this can be used to format DateTime instances or any other objects
+ * that require custom or additional logic when exported as a string.
+ */
+interface DataFormatterInterface
+{
+    /**
+     * Determine if we know how to format this particular data type
+     *
+     * @param $data
+     * @return bool
+     */
+    public function supports($data);
+
+    /**
+     * Formats data
+     *
+     * @param $data
+     * @param PropertyPath $propertyPath
+     * @return string
+     */
+    public function format($data, PropertyPath $propertyPath);
+
+    /**
+     * Returns where in the formatting stack this should land must be greater than 0
+     * Lower priority is processed first. The first formatter found that supports a
+     * particular data type is called and no further processing occurs.
+     *
+     * @return int
+     */
+    public function getPriority();
+}
--- a/src/Formatter/DateTimeFormatter.php
+++ b/src/Formatter/DateTimeFormatter.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Exporter\Formatter;
+
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+final class DateTimeFormatter implements DataFormatterInterface
+{
+    /** @var null|string */
+    private $format = null;
+
+    /**
+     * DateTimeFormatter constructor.
+     * @param null|string $dateTimeFormat
+     */
+    public function __construct($dateTimeFormat = 'r')
+    {
+        $this->format = $dateTimeFormat;
+    }
+
+    /**
+     * @inheritDoc
+     * NEXT_MAJOR: can drop instanceof \DateTime can be removed when dropping php < 5.5
+     */
+    public function supports($data)
+    {
+        return $data instanceof \DateTime || $data instanceof \DateTimeInterface;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function format($data, PropertyPath $propertyPath)
+    {
+        return $data->format($this->format);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getPriority()
+    {
+        return 100;
+    }
+}
--- a/src/Formatter/ObjectToStringFormatter.php
+++ b/src/Formatter/ObjectToStringFormatter.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Exporter\Formatter;
+
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+final class ObjectToStringFormatter implements DataFormatterInterface
+{
+    /**
+     * @inheritDoc
+     */
+    public function supports($data)
+    {
+        return is_object($data);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function format($data, PropertyPath $propertyPath)
+    {
+        return (string) $data;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getPriority()
+    {
+        return 200;
+    }
+}
--- a/src/Source/DoctrineODMQuerySourceIterator.php
+++ b/src/Source/DoctrineODMQuerySourceIterator.php
@@ -14,6 +14,10 @@ namespace Exporter\Source;
 use Doctrine\ODM\MongoDB\Query\Query;
 use Doctrine\ORM\Internal\Hydration\IterableResult;
 use Exporter\Exception\InvalidMethodCallException;
+use Exporter\Formatter\ArrayFormatter;
+use Exporter\Formatter\DataFormatterInterface;
+use Exporter\Formatter\DateTimeFormatter;
+use Exporter\Formatter\ObjectToStringFormatter;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyPath;
 
@@ -40,14 +44,23 @@ class DoctrineODMQuerySourceIterator implements SourceIteratorInterface
     protected $propertyAccessor;
 
     /**
-     * @var string default DateTime format
+     * @var string $dateTimeFormat
      */
     protected $dateTimeFormat;
 
     /**
-     * @param Query  $query          The Doctrine Query
-     * @param array  $fields         Fields to export
-     * @param string $dateTimeFormat
+     * @var DataFormatterInterface[]
+     */
+    private $formatters = array();
+
+    /**
+     * NEXT_MAJOR: Change function signature to __construct(Query $query, array $fields, array $formatters)
+     *             Remove default set of formatters
+     *
+     * @param Query                    $query          The Doctrine Query
+     * @param array                    $fields         Fields to export
+     * @param string                   $dateTimeFormat
+     * @param DataFormatterInterface[] $formatters     Array of data formatters
      */
     public function __construct(Query $query, array $fields, $dateTimeFormat = 'r')
     {
@@ -64,7 +77,29 @@ class DoctrineODMQuerySourceIterator implements SourceIteratorInterface
             }
         }
 
-        $this->dateTimeFormat = $dateTimeFormat;
+        if (func_num_args() == 3 && is_string($dateTimeFormat)) {
+            // When removed, the code below initializing a default set of formatters should be rewritten as well
+            @trigger_error('Passing a dateTimeFormat string as the 4th parameter is deprecated since 1.7.2, to be removed in 2.0. '.
+                'Pass an array of DataFormatterInterfaces instead',
+                E_USER_DEPRECATED);
+        }
+
+        $formatters = array(new ArrayFormatter(), new DateTimeFormatter($dateTimeFormat), new ObjectToStringFormatter());
+        if (func_num_args() == 4) {
+            $formatters = array_merge($formatters, (array)func_get_arg(3));
+        }
+
+        foreach ($formatters as $formatter) {
+            $this->addFormatter($formatter);
+        }
+    }
+
+    /**
+     * @param DataFormatterInterface $formatter
+     */
+    private function addFormatter(DataFormatterInterface $formatter)
+    {
+        $this->formatters[$formatter->getPriority()] = $formatter;
     }
 
     /**
@@ -77,7 +112,7 @@ class DoctrineODMQuerySourceIterator implements SourceIteratorInterface
         $data = [];
 
         foreach ($this->propertyPaths as $name => $propertyPath) {
-            $data[$name] = $this->getValue($this->propertyAccessor->getValue($current, $propertyPath));
+            $data[$name] = $this->getValue($current, $propertyPath);
         }
 
         $this->query->getDocumentManager()->getUnitOfWork()->detach($current);
@@ -118,6 +153,8 @@ class DoctrineODMQuerySourceIterator implements SourceIteratorInterface
             throw new InvalidMethodCallException('Cannot rewind a Doctrine\ODM\Query');
         }
 
+        ksort($this->formatters);
+
         $this->iterator = $this->query->iterate();
         $this->iterator->rewind();
     }
@@ -127,11 +164,20 @@ class DoctrineODMQuerySourceIterator implements SourceIteratorInterface
      */
     public function setDateTimeFormat($dateTimeFormat)
     {
+        @trigger_error('Passing a dateTimeFormat string is deprecated since 1.7.2, to be removed in 2.0. '.
+            'The source should be initialized in the constructor with an array of DataFormatterInterfaces instead',
+            E_USER_DEPRECATED);
         $this->dateTimeFormat = $dateTimeFormat;
+
+        $newFormatter = new DateTimeFormatter($dateTimeFormat);
+        $this->formatters[$newFormatter->getPriority()] = $newFormatter;
     }
 
     /**
+     * NEXT_MAJOR: Remove
+     *
      * @return string
+     * @deprecated Deprecated since 1.7.1, to be removed in 2.0
      */
     public function getDateTimeFormat()
     {
@@ -139,18 +185,32 @@ class DoctrineODMQuerySourceIterator implements SourceIteratorInterface
     }
 
     /**
-     * @param $value
+     * NEXT_MAJOR: Change function signature to getValue($target, PropertyPath $propertyPath), keep code between if block,
+     *             and drop the else code block
      *
-     * @return null|string
+     * @param mixed        $target
+     * @param PropertyPath $propertyPath
+     *
+     * @return mixed|array|string|object
      */
-    protected function getValue($value)
+    protected function getValue($target)
     {
-        if (is_array($value) || $value instanceof \Traversable) {
-            $value = null;
-        } elseif ($value instanceof \DateTimeInterface) {
-            $value = $value->format($this->dateTimeFormat);
-        } elseif (is_object($value)) {
-            $value = (string) $value;
+        if (func_num_args() == 2 && func_get_arg(1) instanceof PropertyPath) {
+            $propertyPath = func_get_arg(1);
+            $value = $this->propertyAccessor->getValue($target, $propertyPath);
+        } else {
+            @trigger_error('Passing a single value is deprecated since 1.7.2, to be removed in 2.0.'.
+                'Pass the target object and a PropertyPath object that will retrieve the value instead',
+                E_USER_DEPRECATED);
+
+            $value = $target;
+            $propertyPath = new PropertyPath('');
+        }
+
+        foreach ($this->formatters as $formatter) {
+            if ($formatter->supports($value)) {
+                return $formatter->format($value, $propertyPath);
+            }
         }
 
         return $value;
--- a/src/Source/DoctrineORMQuerySourceIterator.php
+++ b/src/Source/DoctrineORMQuerySourceIterator.php
@@ -13,6 +13,10 @@ namespace Exporter\Source;
 
 use Doctrine\ORM\Query;
 use Exporter\Exception\InvalidMethodCallException;
+use Exporter\Formatter\ArrayFormatter;
+use Exporter\Formatter\DataFormatterInterface;
+use Exporter\Formatter\DateTimeFormatter;
+use Exporter\Formatter\ObjectToStringFormatter;
 use Symfony\Component\PropertyAccess\Exception\UnexpectedTypeException;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyAccessor;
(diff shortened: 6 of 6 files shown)


src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -139,20 +144,52 @@ public function getDateTimeFormat()
...
     {
+        $value = $this->propertyAccessor->getValue($target, $propertyPath);
         if (is_array($value) || $value instanceof \Traversable) {
             $value = null;
         } elseif ($value instanceof \DateTime) {
-            $value = $value->format($this->dateTimeFormat);
+            $value = $this->getFormattedDateForField($value, $propertyPath);
         } elseif (is_object($value)) {
             $value = (string) $value;
         }
```
greg0ire: This if… elseif… elseif… chain is an OCP violation. There should be a new interface called SomethingFormatter, with one implementation per type and a default implementation.
Could look like

```
interface SomethingFormatterInterface
{
    public function supports($data);
    public function format($value, $propertyPath);
}
```

That would remove a lot of code duplication. If you do this, then I think your idea can also be implemented afterwards.

gnat42: I was just thinking the same thing as I looked at using this within my own project. What namespace would you want the Interface? Exporter\Formatter?

greg0ire: Yes, it would make sense.

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -45,6 +45,11 @@ class DoctrineODMQuerySourceIterator implements SourceIteratorInterface
     protected $dateTimeFormat;
 
     /**
+     * @var array
```
core23: Maybe `@var string[]`?

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -139,20 +144,52 @@ public function getDateTimeFormat()
     }
 
     /**
-     * @param $value
+     * @return array
+     */
+    public function getDateFormats()
```
core23: final

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -139,20 +144,52 @@ public function getDateTimeFormat()
...
+    {
+        return $this->dateFormats;
+    }
+
+    /**
+     * @param array $dateFormats
+     */
+    public function setDateFormats(array $dateFormats = array())
```
core23: final

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -139,20 +144,52 @@ public function getDateTimeFormat()
...
+
+    /**
+     * @param mixed        $target
+     * @param PropertyPath $propertyPath
      *
-     * @return null|string
+     * @return mixed
      */
-    protected function getValue($value)
+    protected function getValue($target, PropertyPath $propertyPath)
```
core23: BC break

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -139,20 +144,52 @@ public function getDateTimeFormat()
...
 
         return $value;
     }
+
+    /**
+     * @param \DateTime    $value
+     * @param PropertyPath $propertyPath
+     * @return string
+     */
+    protected function getDateFormatForField(\DateTime $value, PropertyPath $propertyPath)
```
core23: final

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -45,6 +45,11 @@ class DoctrineORMQuerySourceIterator implements SourceIteratorInterface
     protected $dateTimeFormat;
 
     /**
+     * @var array
+     */
+    protected $dateFormats = array();
```
core23: private?

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -45,6 +45,11 @@ class DoctrineODMQuerySourceIterator implements SourceIteratorInterface
     protected $dateTimeFormat;
 
     /**
+     * @var array
+     */
+    protected $dateFormats = array();
```
core23: private?

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -147,20 +152,52 @@ public function getDateTimeFormat()
...
+     */
+    public function getDateFormats()
+    {
+        return $this->dateFormats;
+    }
+
+    /**
+     * @param array $dateFormats
+     */
+    public function setDateFormats(array $dateFormats = array())
```
core23: final

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -147,20 +152,52 @@ public function getDateTimeFormat()
...
+
+    /**
+     * @param mixed        $target
+     * @param PropertyPath $propertyPath
      *
-     * @return null|string
+     * @return mixed
      */
-    protected function getValue($value)
+    protected function getValue($target, PropertyPath $propertyPath)
```
core23: BC break

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -147,20 +152,52 @@ public function getDateTimeFormat()
...
 
         return $value;
     }
+
+    /**
+     * @param \DateTime    $value
+     * @param PropertyPath $propertyPath
+     * @return string
+     */
+    protected function getDateFormatForField(\DateTime $value, PropertyPath $propertyPath)
```
core23: final

src/Source/DoctrineORMQuerySourceIterator.php
gnat42: I went with protected since the $dateTimeFormat was also protected

src/Source/DoctrineODMQuerySourceIterator.php
gnat42: So this change is a major BC break so I would need to base it off master I presume?

src/Source/DoctrineORMQuerySourceIterator.php
greg0ire: We are using private for new things now, because it's easier to change w/o a BC-break

gnat42:
Ok @greg0ire how about like this? Since it was a major BC break I didn't leave the $dateTimeFormat argument but for both the ORM/ODM sources the constructor could be this instead.

```
/**
     * @param Query                    $query          The Doctrine Query
     * @param array                    $fields         Fields to export
     * @param string                   $dateTimeFormat
     * @param DataFormatterInterface[] $formatters     Array of data formatters
     */
    public function __construct(Query $query, array $fields, $dateTimeFormat = 'r', array $formatters = array())
    {
        $this->query = clone $query;

        $this->propertyAccessor = PropertyAccess::createPropertyAccessor();

        $this->propertyPaths = array();
        foreach ($fields as $name => $field) {
            if (is_string($name) && is_string($field)) {
                $this->propertyPaths[$name] = new PropertyPath($field);
            } else {
                $this->propertyPaths[$field] = new PropertyPath($field);
            }
        }

        if (empty($formatters)) {
            $this->formatters = array(new ArrayFormatter(), new DateTimeFormatter($dateTimeFormat), new ObjectToStringFormatter());
        } else {
            foreach($formatters as $formatter) {
                if(!$formatter instanceof DataFormatterInterface) {
                    throw new InvalidArgumentException('Formatters must implement '.DataFormatterInterface::class);
                }
                $this->formatters[] = $formatter;
            }
        }
    }
```

Where we still allow to pass a default DateTimeFormat that would get passed to the default set of formatters. Thoughts?

greg0ire:
I think the iterators should know about date time formats. Not that you can keep `__construct` as is and use `func_get_args()` if you need to simulate a different signature, it will be BC.

gnat42:
Well the iterators know about the date time formats, but by default I'm making it basically work the way it used to but allowing a method to change behaviour. So currently the exporter uses 'r' as the default DateTime format. This way that still occurs. If however someone wants something more complex they would pass their own array of formatters. 

The only issue I can see is that as a user, I wouldn't know I needed the array and object transformers in the position they find themselves. So I wonder if I need more logic in the constructor if they pass an array to insert theirs in between but replacing the date one if the user passed one of those in. Or if this can simply be documented?

gnat42:
Also its already a BC because I changed the getValue function signature. I thought I could do this without any BCs because it was a protected function, but well @core23 rightfully pointed out that that's a BC since someone could be extending these objects and overriding those functions. So this is a BC regardless. The question then is what should the constructor *be*.

greg0ire:
My comment about `func_get_args()` applies to the protected method BC too.

gnat42:
ah - I get it @greg0ire ...

greg0ire:
IMO you could add a `getPriority()` method to the `WhateverFormatterInterface`, that will solve your problem.

src/Formatter/ArrayFormatter.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Exporter\Formatter;
+
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+class ArrayFormatter implements DataFormatterInterface
```
core23: Final for all new classes and/or methods please

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -40,16 +45,16 @@ class DoctrineODMQuerySourceIterator implements SourceIteratorInterface
...
     /**
-     * @param Query  $query          The Doctrine Query
-     * @param array  $fields         Fields to export
-     * @param string $dateTimeFormat
+     * @param Query                    $query          The Doctrine Query
+     * @param array                    $fields         Fields to export
+     * @param DataFormatterInterface[] $formatters     Array of data formatters
      */
-    public function __construct(Query $query, array $fields, $dateTimeFormat = 'r')
+    public function __construct(Query $query, array $fields, array $formatters = array())
```
core23: BC break, because the method signature is changed

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -123,34 +137,19 @@ public function rewind()
     }
 
     /**
-     * @param string $dateTimeFormat
-     */
-    public function setDateTimeFormat($dateTimeFormat)
```
core23: Removing public methods is a BC break

gnat42:
OK, re-implemented using func_get_args and friends. Thoughts?

src/Exception/InvalidArgumentException.php
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Exporter\Exception;
+
+class InvalidArgumentException extends \InvalidArgumentException
```
core23: New class => final

src/Formatter/DataFormatterInterface.php
```diff
@@ -0,0 +1,45 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Exporter\Formatter;
+
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+/**
+ * Interface DataFormatterInterface
+ * @package Exporter\Formatter
```
core23: We don't need the `@package` anotation here

src/Formatter/DataFormatterInterface.php
```diff
@@ -0,0 +1,45 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Exporter\Formatter;
+
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+/**
+ * Interface DataFormatterInterface
```
core23: Can you just say something more then repeating the interface name

src/Formatter/DataFormatterInterface.php
```diff
@@ -0,0 +1,45 @@
...
+ * @package Exporter\Formatter
+ */
```
greg0ire: Please remove this

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -40,14 +45,15 @@ class DoctrineODMQuerySourceIterator implements SourceIteratorInterface
     protected $propertyAccessor;
 
     /**
-     * @var string default DateTime format
+     * @var DataFormatterInterface[]
      */
-    protected $dateTimeFormat;
```
core23: Removing a protected / public property / method is a BC break

src/Formatter/DataFormatterInterface.php
```diff
@@ -0,0 +1,45 @@
...
+     * Formats data
+     *
+     * @param $data
+     * @param PropertyPath $propertyPath
+     * @return string
+     */
+    public function format($data, PropertyPath $propertyPath);
+
+    /**
+     * Returns where in the formatting stack this should land must be greater than 0
```
greg0ire: Please document what a low value means

src/Formatter/DateTimeFormatter.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return $data->format($this->dateTimeFormat);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getPriority()
+    {
+        return PHP_INT_MAX-2;
```
greg0ire: Please make room for other formatters, by using 10, 20, 30, instead of 1, 2, 3, and maybe PHP_INT_MAX is overkill :laughing:

src/Formatter/ObjectToStringFormatter.php
```diff
@@ -0,0 +1,41 @@
...
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+final class ObjectToStringFormatter implements DataFormatterInterface
+{
+    /**
+     * @inheritDoc
+     */
+    public function supports($data)
+    {
+        return (is_object($data));
```
greg0ire: Please remove unneeded parenthesis

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -64,7 +70,23 @@ public function __construct(Query $query, array $fields, $dateTimeFormat = 'r')
             }
         }
 
-        $this->dateTimeFormat = $dateTimeFormat;
+        $formatters = array(new ArrayFormatter(), new DateTimeFormatter($dateTimeFormat), new ObjectToStringFormatter());
```
greg0ire: Please put this in an else clause and trigger a deprecation error, and ask people to remove this on next major

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -135,30 +158,30 @@ public function rewind()
...
-            $value = $value->format($this->dateTimeFormat);
-        } elseif (is_object($value)) {
-            $value = (string) $value;
+        if (func_num_args() == 2 && func_get_arg(1) instanceof PropertyPath) {
+            $propertyPath = func_get_arg(1);
+            $value = $this->propertyAccessor->getValue($target, $propertyPath);
+        } else {
+            $value = $target;
+            $propertyPath = new PropertyPath('');
+        }
```
greg0ire: Please add deprecations here too

src/Source/DoctrineODMQuerySourceIterator.php
gnat42: Well, why is this deprecated? This particular set of code can stay as is if you ask me. If you don't require per property DateTime formats it should be fine?

greg0ire: Because it's a DIC violation. Dependencies should be injected. It's also an OCP violation. If someday we deem a new formatter is worth adding by default, we should't edit this file, we should just create a new service.

gnat42:
to add a deprecation you want /** @deprecation('message') */ inline in the code? or @trigger_error(message) ? Also what version does this target as removal then?

greg0ire:
@gnat42 please go the `trigger_error` way, as documented in CONTRIBUTING.md, and target the next major version

gnat42:
Alrighty... how about now?

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -64,7 +75,30 @@ public function __construct(Query $query, array $fields, $dateTimeFormat = 'r')
...
+            @trigger_error('Passing a dateTimeFormat string is deprecated since 1.7.2, to be removed in 2.0. '.
+                'Pass an array of DataFormatterInterfaces instead',
+                E_USER_DEPRECATED);
+        }
+
+        $formatters = array(new ArrayFormatter(), new DateTimeFormatter($dateTimeFormat), new ObjectToStringFormatter());
+        if (func_num_args() == 4) {
+            $additionalFormatters = func_get_arg(4);
+            if(!is_array($additionalFormatters)) {
+                throw new InvalidArgumentException('4th parameter must be an array of ' . DataFormatterInterface::class . ' classes');
```
greg0ire: You can't use `::class` yet, sorry

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -64,7 +75,30 @@ public function __construct(Query $query, array $fields, $dateTimeFormat = 'r')
             }
         }
 
-        $this->dateTimeFormat = $dateTimeFormat;
+        if (func_num_args() == 3 && is_string($dateTimeFormat)) {
+            // When removed, the code below initializing a default set of formatters should be rewritten as well
+            @trigger_error('Passing a dateTimeFormat string is deprecated since 1.7.2, to be removed in 2.0. '.
+                'Pass an array of DataFormatterInterfaces instead',
```
greg0ire: "as a fourth argument", let's make it obvious, b/c the code isn't :P

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -64,7 +75,30 @@ public function __construct(Query $query, array $fields, $dateTimeFormat = 'r')
...
+            $additionalFormatters = func_get_arg(4);
+            if(!is_array($additionalFormatters)) {
+                throw new InvalidArgumentException('4th parameter must be an array of ' . DataFormatterInterface::class . ' classes');
+            }
+
+            $formatters = array_merge($formatters, $additionalFormatters);
+        }
+
+        foreach($formatters as $formatter) {
+            if(!$formatter instanceof DataFormatterInterface) {
```
greg0ire: Introduce a private method with this as type hinting and you won't need to throw a custom exception.

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -127,30 +161,38 @@ public function rewind()
...
-        $this->dateTimeFormat = $dateTimeFormat;
-    }
+        @trigger_error('Passing a dateTimeFormat string is deprecated since 1.7.2, to be removed in 2.0. '.
+            'The source should be initialized in the constructor with an array of DataFormatterInterfaces instead',
+            E_USER_DEPRECATED);
 
-    /**
-     * @return string
-     */
-    public function getDateTimeFormat()
```
greg0ire: Removing a method is a BC-break

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -127,30 +161,38 @@ public function rewind()
...
+        $this->formatters[$newFormatter->getPriority()] = $newFormatter;
     }
 
     /**
-     * @param $value
+     * @param mixed        $target
+     * @param PropertyPath $propertyPath
      *
-     * @return null|string
+     * @return mixed
```
greg0ire: mixed what? Please list possible types separated with `|` (keep the `mixed` though)

src/Formatter/DateTimeFormatter.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $this->dateTimeFormat = $dateTimeFormat;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function supports($data)
+    {
+        return ($data instanceof \DateTime);
```
greg0ire: - please remove uneeded parenthesis
- `$data instanceof \DateTime || $data instanceof \DateTimeInterface` (the left part can be dropped when dropping php < 5.5)

src/Source/DoctrineODMQuerySourceIterator.php
```diff
@@ -40,14 +45,20 @@ class DoctrineODMQuerySourceIterator implements SourceIteratorInterface
...
      */
     protected $dateTimeFormat;
 
     /**
-     * @param Query  $query          The Doctrine Query
-     * @param array  $fields         Fields to export
-     * @param string $dateTimeFormat
+     * @var DataFormatterInterface[]
+     */
+    protected $formatters = array();
```
greg0ire: private please

src/Source/DoctrineODMQuerySourceIterator.php
greg0ire: You can also make `$this->formatters` [a priority queue](http://php.net/manual/en/class.splpriorityqueue.php) if you think it isn't overkill

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -135,30 +170,38 @@ public function rewind()
...
-    protected function getValue($value)
+    protected function getValue($target)
     {
-        if (is_array($value) || $value instanceof \Traversable) {
-            $value = null;
-        } elseif ($value instanceof \DateTime) {
-            $value = $value->format($this->dateTimeFormat);
-        } elseif (is_object($value)) {
-            $value = (string) $value;
+        if (func_num_args() == 2 && func_get_arg(1) instanceof PropertyPath) {
```
greg0ire: Please add `NEXT_MAJOR` instructions where needed, as described in CONTRIBUTING.md

src/Source/DoctrineODMQuerySourceIterator.php
gnat42: I don't think it needs a priority queue - Mainly because this really should be a very simple ordered array, and when using xdebug, a priority queue is opaque so for anyone debugging for whatever reason its a huge pain.

greg0ire: As you wish :+1:

gnat42:
Alrighty - I think I've completed the remaining requests

src/Formatter/DateTimeFormatter.php
```diff
@@ -0,0 +1,54 @@
...
+     * @param null|string $dateTimeFormat
+     */
+    public function __construct($dateTimeFormat = 'r')
+    {
+        $this->dateTimeFormat = $dateTimeFormat;
+    }
+
+    /**
+     * @inheritDoc
+     * NEXT_MAJOR: can drop instanceof \DateTimeInterface can be removed when dropping php < 5.5
```
greg0ire: No, it's `instanceof \DateTime` which should be dropped.

gnat42:
updated...

gnat42:
So a simple doc on how these two sources can use it? Since it wasn't added to any other sources.

greg0ire:
I think the docs should explain how one can add a custom formatter.

greg0ire:
A compiler pass [similar to this one](https://github.com/sonata-project/exporter/blob/1.x/src/Bridge/Symfony/DependencyInjection/Compiler/ExporterCompilerPass.php#L39) would be nice to do the job, BTW

gnat42:
So I would love to be able to add docs, but am out of time on this and have to move on to other tasks.

greg0ire:
> So I would love to be able to add docs, but am out of time on this and have to move on to other tasks.

You mean you're never gonna do it, or just you're gonna do it later? Because we can totally wait for the docs, but will probably not merge without them : people keep complaining about how the Sonata docs are not great.

greg0ire:
Same goes for tests BTW. I might be able to help you finish this, but… time is also a limitation for me.

gnat42:
I'll see if I can add some (of both) maybe Saturday.

src/Formatter/DateTimeFormatter.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+namespace Exporter\Formatter;
+
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+final class DateTimeFormatter implements DataFormatterInterface
+{
+    /** @var null|string */
+    private $dateTimeFormat = null;
```
OskarStark: isn't `$format` enough, instead of `$dateTimeFormat` @greg0ire ?

OskarStark: ah we use it here already, too

https://github.com/sonata-project/exporter/pull/147/files#diff-59abbb9207041c402aca780bf90bc52fR47

so please ignore my comment

greg0ire: You're right, it's a bit repetitive

greg0ire: It's a private property now, can't be used outside the class, so we can name it Oskar if we please

OskarStark:
ping @gnat42

gnat42:
Right... I wrote up some simple tests for the interface and that's about it so far. I'll see about getting a few more tests and some docs this week. I've been using my fork of this with this PR since. It works wonderfully.

gnat42:
Honestly, at this point 1.5 years later I don't even know what project I needed this for and I probably have it in a fork or something. I don't use Sonata much so feel free to rebase or whatever. I don't even remember what branch it was targeting and which it should now. If you want the change feel free to re-implement this in whatever branch you need.

greg0ire:
Ok too bad… thanks anyway. Let's close, people that need this can find it and carry it if they want to.

PR #148, opened 2017-03-20
Fixed correct classname in README and add missing use statements

eko:
## Subject

Fixed correct classname in README and add missing use statements.


eko:
@greg0ire You're right, I've fixed the commit message and pull request title.

jordisala1991:
Thanks @eko

greg0ire:
Thanks @eko , next time consider even doing 2 commits (with "git add -p"), if you have "and" in your commit message, it's a smell.

PR #151, opened 2017-05-16
Update symfony.rst

carlos-granados:
I am targeting this branch, because this is backwards compatible.

### Fixed
```
docs/reference/symfony.rst
```
## Subject

The register bundle section has an error in the code. Since this is registered within an array, the line needs to end with `,` not `;`


greg0ire:
Thanks @carlos-granados !

greg0ire:
I removed the changelog, which is not needed for docs PRs

PR #180, opened 2017-10-15
Add void return type hint

greg0ire:
## Changelog

```markdown
### Added
- `void` return type hint wherever applicable

```

### DIFF
--- a/src/Exporter.php
+++ b/src/Exporter.php
@@ -57,7 +57,7 @@ final class Exporter
         }
         $writer = $this->writers[$format];
 
-        $callback = function () use ($source, $writer) {
+        $callback = function () use ($source, $writer): void {
             $handler = \Exporter\Handler::create($source, $writer);
             $handler->export();
         };
@@ -86,7 +86,7 @@ final class Exporter
      *
      * @param TypedWriterInterface $writer a possible writer for exporting data
      */
-    public function addWriter(TypedWriterInterface $writer)
+    public function addWriter(TypedWriterInterface $writer): void
     {
         $this->writers[$writer->getFormat()] = $writer;
     }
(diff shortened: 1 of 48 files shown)


src/Exporter.php
```diff
@@ -57,7 +57,7 @@ public function getResponse(string $format, string $filename, SourceIteratorInte
         }
         $writer = $this->writers[$format];
 
-        $callback = function () use ($source, $writer) {
+        $callback = function () use ($source, $writer): void {
```
core23: This is needed? oO

greg0ire: If you mean "this works?", yes it works :) (why not?). If you mean "Is this necessary?" none of this PR is, it just makes things safer.

core23: I can understand this stuff for the api methods (protected and public things). But for internal functions? I don't think that this is useful at all.

greg0ire: It doesn't hurt though, and I use php-cs-fixer for this, and I'd hate to make an exception for this kind of stuff each time I run it. Also I'd argue it's good to have type hinting for internal functions: it helps your IDE understanding what happens, and if you do wrong things it crashes early.

tests/ExporterTest.php
```diff
@@ -30,7 +30,7 @@ public function testFilter()
         $exporter->getResponse('foo', 'foo', $source);
     }
 
-    public function testConstructorRejectsNonTypedWriters()
+    public function testConstructorRejectsNonTypedWriters(): void
     {
         $this->setExpectedException(
             version_compare(PHP_VERSION, '7.0.0', '<') ? 'PHPUnit_Framework_Error' : 'TypeError',
```
jordisala1991: PHP5 line detected!

greg0ire: @jordisala1991 fixed

greg0ire:
@sonata-project/contributors please merge

PR #185, opened 2017-10-23
Phpstan

greg0ire:
I am targeting this branch, because this should not be backported

## Changelog

Not sure if it is useful, especially for the bug since there is no major release yet

```markdown
### Removed
- support for phpunit 4 in public test

### Fixed
- crash when using the exporter class
```

### DIFF
--- a/src/Exporter.php
+++ b/src/Exporter.php
@@ -58,7 +58,7 @@ final class Exporter
         $writer = $this->writers[$format];
 
         $callback = function () use ($source, $writer): void {
-            $handler = \Exporter\Handler::create($source, $writer);
+            $handler = Handler::create($source, $writer);
             $handler->export();
         };
 
--- a/src/Test/AbstractTypedWriterTestCase.php
+++ b/src/Test/AbstractTypedWriterTestCase.php
@@ -11,12 +11,13 @@
 
 namespace Sonata\Exporter\Test;
 
+use PHPUnit\Framework\TestCase;
 use Sonata\Exporter\Writer\TypedWriterInterface;
 
 /**
  * @author Grégoire Paris <postmaster@greg0ire.fr>
  */
-abstract class AbstractTypedWriterTestCase extends \PHPUnit_Framework_TestCase
+abstract class AbstractTypedWriterTestCase extends TestCase
 {
     /**
      * @var WriterInterface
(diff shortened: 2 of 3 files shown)


Soullivaneuh:
Could you please be a little bit more explicit about what you fixed on the changelog?

greg0ire:
Done

Soullivaneuh:
Can a simple test be written just to cover this big issue?

greg0ire:
No.

greg0ire:
Kidding, here you go!

greg0ire:
@core23 dropping phpunit 4 in a public test on stable is an unnecessary BC-break. So is the new test if we release a new major soon.

core23:
> @core23 dropping phpunit 4 in a public test on stable is an unnecessary BC-break. So is the new test if we release a new major soon.

We are talking about tests... Also PHPUnit was never a (required) dependency with a version notation. In this case we could live with a BC break.

greg0ire:
>  Also PHPUnit was never a (required) dependency with a version notation.

That's exactly why this is a BC-break.

Soullivaneuh:
@greg0ire is right, we are talking about a **reusable** test case here. Changing the namespace for class extends is BC break.

src/Exporter.php
```diff
@@ -58,7 +58,7 @@ public function getResponse(string $format, string $filename, SourceIteratorInte
         $writer = $this->writers[$format];
 
         $callback = function () use ($source, $writer): void {
-            $handler = \Exporter\Handler::create($source, $writer);
```
jordisala1991: Care with this. Handler namespace is different From Exporter namespace

greg0ire: No it isn't they are both at the root of `src`

jordisala1991: Ohhh it was different on 1.x not on master anymore

greg0ire: ? they are both in the same namespace on 1.x too? 🤔

jordisala1991: Yes.. indeed. But I was looking at your Exporter class namespace from master vs Handler namespace from 1.x and it were different. :)

greg0ire: Oooh ok alright then :)

greg0ire:
How about merging @sonata-project/contributors ?

PR #199, opened 2017-11-30
Replace type checking of DateTime with DateTimeInterface

jaikdean:
I am targeting this branch, because there are no BC breaks.

## Changelog

```markdown
### Fixed
- Allow `\DateTimeImmutable` values
```

## Subject

This PR changes value type checking from `instanceof \DateTime` to `instanceof \DateTimeInterface`, which allows `\DateTimeImmutable` values to be exported correctly. Previously, this would cause a fatal error "Object of class DateTimeImmutable could not be converted to string at vendor/sonata-project/exporter/src/Source/DoctrineORMQuerySourceIterator.php:161)".


greg0ire:
Thanks @jaikdean !

PR #208, opened 2017-12-10
Fix coding standards

jordisala1991:
After dev-kit PR https://github.com/sonata-project/exporter/pull/207, this fixed the build

### DIFF
--- a/src/Writer/SitemapWriter.php
+++ b/src/Writer/SitemapWriter.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /*
  * This file is part of the Sonata Project package.
  *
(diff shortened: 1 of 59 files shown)


src/Writer/SitemapWriter.php
```diff
@@ -16,8 +18,8 @@
  */
 class SitemapWriter implements WriterInterface
 {
-    const LIMIT_SIZE = 10485760;
-    const LIMIT_URL = 50000;
+    public const LIMIT_SIZE = 10485760;
```
core23: Can we make this private?

jordisala1991: That change from public to private would be BC break, but since we are on master it can be done.

I am okay with it, but would prefer to do it in another PR, because that change comes from simply execute `php-cs-fixer`.

PR #212, opened 2018-01-21
Applied fixes from FlintCI

Soullivaneuh:


src/Source/XmlSourceIterator.php
```diff
@@ -90,7 +90,6 @@ public function tagEnd($parser, $name)
     public function tagContent($parser, $data)
     {
         if (isset($this->bufferedRow['i_'.$this->currentRowIndex], $this->bufferedRow['i_'.$this->currentRowIndex][$this->currentColumnIndex])
-
```
OskarStark: This looks strange

jordisala1991: It is normal there was a extra line : https://github.com/sonata-project/exporter/blob/1.x/src/Source/XmlSourceIterator.php#L93

OskarStark: hmm, but shouldn't the closing bracket of the `if` be on the same line instead of `) {` ?

PR #220, opened 2018-04-30
[Improvement] Add custom escape/terminator to CsvWriter

### LINKED ISSUE #219: [Bug] Actually use the escape parameter.
## Subject
The construct method takes an escape character string, but never actually uses it when running fputcsv. 

## Steps to reproduce
Try using CsvWriter with a custom escape character.

## Expected results
You'd expect that character to be used, since it's taken as a param.

## Actual results
It uses the default escape character always.

heathdutton:
I am targeting this branch, because this is not BC breaking, just using the existing parameter and adding a new one (which is optional) for CSV generation to be more backward compatible with old formats.

Closes #219

### Added
- Add custom escape/terminator strings to `CsvWriter`.


CHANGELOG.md
```diff
@@ -46,3 +46,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [1.2.2](https://github.com/sonata-project/exporter/compare/1.2.1...1.2.2) - 2013-05-02
 ### Added
 - Add new argument in method \Exporter\Writer\SitemapWriter::generateSitemapIndex to handle absolute URL.
+
+### Added
```
kunicmarko20: This edits sonata maintainer, please remove this edit

src/Writer/CsvWriter.php
```diff
@@ -58,21 +58,28 @@ class CsvWriter implements TypedWriterInterface
...
      * @param string $filename
      * @param string $delimiter
      * @param string $enclosure
      * @param string $escape
      * @param bool   $showHeaders
      * @param bool   $withBom
+     * @param string $terminate
      */
-    public function __construct($filename, $delimiter = ',', $enclosure = '"', $escape = '\\', $showHeaders = true, $withBom = false)
+    public function __construct(string $filename, string $delimiter = ',', string $enclosure = '"', string $escape = '\\', bool $showHeaders = true, bool $withBom = false, string $terminate = "\n")
```
kunicmarko20: We support php 5.6, please remove string typehints

kunicmarko20: Bool also

kunicmarko20: Also, this line is longer than 120 chars, please break it

src/Writer/CsvWriter.php
```diff
@@ -58,21 +58,28 @@ class CsvWriter implements TypedWriterInterface
...
     {
         $this->filename = $filename;
         $this->delimiter = $delimiter;
         $this->enclosure = $enclosure;
         $this->escape = $escape;
         $this->showHeaders = $showHeaders;
+        $this->terminate = $terminate;
```
kunicmarko20: You are talking about how eacape parameter is not used but you are also adding terminate, you will have to do this in a separate PR. 1 PR = 1 fix/feature

CHANGELOG.md
heathdutton: :+1:

src/Writer/CsvWriter.php
heathdutton: :+1:

src/Writer/CsvWriter.php
heathdutton: Will do.

heathdutton:
Will recreate.

PR #221, opened 2018-04-30
[Bug] Actually use the escape parameter.

### LINKED ISSUE #219: [Bug] Actually use the escape parameter.

heathdutton:
I am targeting this branch, because this is not BC breaking, just a breakfix.

Closes #219 

### Fixed
- CsvWriter actually uses the escape parameter.

## Subject

Simplified version of the PR to just fix the missing params.

### DIFF
--- a/tests/Writer/CsvWriterTest.php
+++ b/tests/Writer/CsvWriterTest.php
@@ -59,9 +59,24 @@ class CsvWriterTest extends AbstractTypedWriterTestCase
         $this->assertEquals($expected, trim(file_get_contents($this->filename)));
     }
 
+    public function testEscapeFormating()
+    {
+        $writer = new CsvWriter($this->filename, ',', '"', '/', false);
+
+        $writer->open();
+
+        $writer->write(['john','doe', '\\', '/']);
+
+        $writer->close();
+
+        $expected = 'john,doe,\,"/"';
+
+        $this->assertEquals($expected, trim(file_get_contents($this->filename)));
+    }
+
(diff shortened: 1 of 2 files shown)


heathdutton:
Looks like the tests need fixing, on it.

greg0ire:
> just a breakfix.

I hope this is not what it sounds like :sweat_smile:

tests/Writer/CsvWriterTest.php
```diff
@@ -59,9 +59,24 @@ public function testEnclosureFormating()
         $this->assertEquals($expected, trim(file_get_contents($this->filename)));
     }
 
+    public function testEscapeFormating()
+    {
+        $writer = new CsvWriter($this->filename, ',', '"', '/', false);
+
+        $writer->open();
+
+        $writer->write(['john','doe', '\\', '/']);
```
greg0ire: flintci is complaining about the missing space before `'doe'`

greg0ire:
Thanks @heathdutton for fixing that embarrassing mistake :sweat_smile:

heathdutton:
Thanks for not laughing at my PRs :)

kunicmarko20:
@heathdutton thank you for finding and fixing this bug!

PR #238, opened 2018-10-01
Fix Tests

kunicmarko20:
Not sure how is the build even passing, tests had the wrong namespace :man_shrugging:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Sonata\\Exporter\\Test\\": "tests/"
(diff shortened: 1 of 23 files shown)


greg0ire:
> Not sure how is the build even passing

Now it isn't :sweat_smile:

greg0ire:
> Not sure how is the build even passing

Now it isn't :sweat_smile: 

>  tests had the wrong namespace

phpunit does not care about the namespaces of the test class at all, that's why :P

kunicmarko20:
blocked by #239

kunicmarko20:
This should be done

PR #246, opened 2018-11-02
Add support for exporting DateInterval values

### LINKED ISSUE #245: Add support for DateInterval
## Feature Request

When exporting DateInterval values in the DoctrineORMQuerySourceIterator class, no file can be downloaded.

kell4770:
I am targeting this branch, because it is a minor feature.

Closes #245

## Changelog

```markdown
### Changed
- Added support for exporting DateInterval values

```

## To do
- [ ] Update the tests?

## Subject

In order to export DateIntervals, I've added an elseif to the getValue function in DoctrineORMQuerySourceIterator, with a getDuration helper function to return an ISO8601 duration.

### DIFF
--- a/src/Source/DoctrineORMQuerySourceIterator.php
+++ b/src/Source/DoctrineORMQuerySourceIterator.php
@@ -158,10 +158,34 @@ class DoctrineORMQuerySourceIterator implements SourceIteratorInterface
             $value = null;
         } elseif ($value instanceof \DateTimeInterface) {
             $value = $value->format($this->dateTimeFormat);
+        } elseif ($value instanceof \DateInterval) {
+            $value = $this->getDuration($value);
         } elseif (\is_object($value)) {
             $value = (string) $value;
         }
 
         return $value;
+    }
+
+    /**
+     * @param \DateInterval $interval
+     *
+     * @return string An ISO8601 duration
+     */
+    protected function getDuration(\DateInterval $interval)
+    {
+        if ($interval->y > 0) {
+            return sprintf('P%dY', $interval->y);
+        } elseif ($interval->m > 0) {
+            return sprintf('P%dM', $interval->m);
+        } elseif ($interval->d > 0) {
+            return sprintf('P%dD', $interval->d);
+        } elseif ($interval->h > 0) {
+            return sprintf('PT%dH', $interval->h);
+        } elseif ($interval->i > 0) {
+            return sprintf('PT%dM', $interval->i);
+        } else {
+            return sprintf('PT%dS', $interval->s);
+        }
     }
 }
--- a/src/Source/DoctrineORMQuerySourceIterator.php
+++ b/src/Source/DoctrineORMQuerySourceIterator.php
@@ -174,21 +174,21 @@ class DoctrineORMQuerySourceIterator implements SourceIteratorInterface
      */
     public static function getDuration(\DateInterval $interval)
     {
-        $dataPartAttributes = array(
+        $datePartAttributes = [
             'y' => 'Y',
             'm' => 'M',
             'd' => 'D',
-        );
-        $timePartAttributes = array(
+        ];
+        $timePartAttributes = [
             'h' => 'H',
             'i' => 'M',
             's' => 'S',
-        );
+        ];
 
         $datePart = '';
-        foreach ($dataPartAttributes as $datePartAttribute => $dataPartAttributeString) {
+        foreach ($datePartAttributes as $datePartAttribute => $datePartAttributeString) {
             if ($dateInterval->$datePartAttribute !== 0) {
-                $datePart .= $dateInterval->$datePartAttribute.$dataPartAttributeString;
+                $datePart .= ($dateInterval->$datePartAttribute).$datePartAttributeString;
             }
         }
 
@@ -198,7 +198,6 @@ class DoctrineORMQuerySourceIterator implements SourceIteratorInterface
                 $timePart .= $dateInterval->$timePartAttribute.$timePartAttributeString;
             }
         }
-
+        
         return 'P'.$datePart.($timePart !== '' ? 'T'.$timePart : '');
-    }
 }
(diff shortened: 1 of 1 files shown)


OskarStark:
Nice, please add some testcase

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -147,6 +147,45 @@ public function getDateTimeFormat()
         return $this->dateTimeFormat;
     }
 
+    /**
+     * @param \DateInterval $dateInterval
```
kunicmarko20: not needed, please remove it

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -147,6 +147,45 @@ public function getDateTimeFormat()
...
+     * @return string An ISO8601 duration
+     */
+    public static function getDuration(\DateInterval $dateInterval)
+    {
+        $datePartAttributes = [
+            'y' => 'Y',
+            'm' => 'M',
+            'd' => 'D',
+        ];
+        $timePartAttributes = [
```
kunicmarko20: feel like this should be const, same as var before, @sonata-project/contributors wdyt?

src/Source/DoctrineORMQuerySourceIterator.php
kell4770: Removed as per request

kunicmarko20: thank you

src/Source/DoctrineORMQuerySourceIterator.php
kell4770: @OskarStark do I need to address this? There's been no response from @sonata-project/contributers

greg0ire: If it does not need to change, please use a constant.

kell4770: Class Constants added

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -147,6 +158,32 @@ public function getDateTimeFormat()
         return $this->dateTimeFormat;
     }
 
+    /**
+     * @return string An ISO8601 duration
+     */
+    public static function getDuration(\DateInterval $dateInterval)
```
kunicmarko20: why is this static?

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -20,6 +20,17 @@
...
+    const DATE_PARTS = [
+            'y' => 'Y',
+            'm' => 'M',
+            'd' => 'D',
+        ];
+    const TIME_PARTS = [
+            'h' => 'H',
+            'i' => 'M',
+            's' => 'S',
+        ];
```
greg0ire: There is an indention issue here… weird that flintCI doesn't complain…

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -147,6 +158,32 @@ public function getDateTimeFormat()
...
+
+        $timePart = '';
+        foreach (self::TIME_PARTS as $timePartAttribute => $timePartAttributeString) {
+            if ($dateInterval->$timePartAttribute !== 0) {
+                $timePart .= $dateInterval->$timePartAttribute.$timePartAttributeString;
+            }
+        }
+
+        if ('' === $datePart && '' === $timePart) {
+            return 'P0Y';
```
kunicmarko20: what does P0Y mean? should that also be a const?

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -20,6 +20,17 @@
...
+    const DATE_PARTS = [
+        'y' => 'Y',
+        'm' => 'M',
+        'd' => 'D',
+        ];
+    const TIME_PARTS = [
+        'h' => 'H',
+        'i' => 'M',
+        's' => 'S',
+        ];
```
kunicmarko20: bracket indentation isn't good

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -20,6 +20,17 @@
 
 class DoctrineORMQuerySourceIterator implements SourceIteratorInterface
 {
+    const DATE_PARTS = [
+        'y' => 'Y',
+        'm' => 'M',
+        'd' => 'D',
+        ];
```
kunicmarko20: bracket indentation isn't good

greg0ire:
```suggestion
    ];
```

src/Source/DoctrineORMQuerySourceIterator.php
greg0ire:
```suggestion
    ];
```

src/Source/DoctrineORMQuerySourceIterator.php
OskarStark: cc @Soullivaneuh 

but it isn't spotted by our lint-job either 🤔  Maybe we need another config option for PHP-CS-Fixer ?

src/Source/DoctrineORMQuerySourceIterator.php
OskarStark: This looks fine to me, it is a datetime format by php

OskarStark:
If the bracket stuff is the only problem here and our CI isn't complaining about it, lets merge as is and fix it later automatically if we have another fixer option available for such cases

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -147,6 +158,32 @@ public function getDateTimeFormat()
         return $this->dateTimeFormat;
     }
 
+    /**
+     * @return string An ISO8601 duration
+     */
+    public function getDuration(\DateInterval $dateInterval)
```
OskarStark:
```suggestion
    public function getDuration(\DateInterval $interval)
```

but this is an optional change

kell4770: Made the change in latest commit

PR #255, opened 2018-11-27
Final all the things

greg0ire:
I am targeting this branch, because this is a BC-break.

## Changelog

```markdown
### Changed
- Most classes have been made final
```

## Subject

Some classes could not be made final because they are extended.


kunicmarko20:
One more off the list, maybe sonata4 happens in under 3 years :joy:

PR #272, opened 2019-01-01
Add forward-compatibility layer

### LINKED ISSUE #198: Add BC way for namespace changes
The namespace will change with the next major release. We should find a way to inform the user, that the namespace will change.

Maybe we could use something like [EasyAdmin](https://github.com/javiereguiluz/EasyAdminBundle) is doing with the legacy folder

greg0ire:
## Subject

I am targeting this branch, because this is BC.

Fixes #198

## Changelog

```markdown
### Added
- Compatiblity layer for `Sonata`-prefixed namespaces
```

### DIFF
--- a/aliases/Bridge/Symfony/Bundle/SonataExporterBundle.php
+++ b/aliases/Bridge/Symfony/Bundle/SonataExporterBundle.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Exporter\Bridge\Symfony\Bundle;
+
+class_alias(
+    '\Sonata\\'.__NAMESPACE__.'\SonataExporterBundle',
+    __NAMESPACE__.'\SonataExporterBundle'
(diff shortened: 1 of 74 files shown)


aliases/Bridge/Symfony/Bundle/SonataExporterBundle.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        $container->addCompilerPass(new ExporterCompilerPass());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getContainerExtensionClass()
+    {
+        return 'Exporter\Bridge\Symfony\DependencyInjection\SonataExporterExtension';
```
kunicmarko20: can't this be `SonataExporterExtension::class`?

PR #276, opened 2019-01-01
Add forward-compatibility layer

### LINKED ISSUE #198: Add BC way for namespace changes

greg0ire:
## Subject
This makes it possible to use the new namespaces in 1.x, which should make migration way easier for everyone

I am targeting this branch, because this is BC.

Fixes #198

## Changelog

```markdown
### Added
- Forward-compatibility layer with `Sonata`-prefixed namespaces
```

## How can I test this on my project?

```
composer config repositories.greg0ire vcs https://github.com/greg0ire/exporter
composer require sonata-project/exporter "dev-compat_layer as 1.42.0"
```

### DIFF
--- a/src/Exporter.php
+++ b/src/Exporter.php
@@ -9,10 +9,10 @@
  * file that was distributed with this source code.
  */
 
-namespace Exporter;
+namespace Sonata\Exporter;
 
-use Exporter\Source\SourceIteratorInterface;
-use Exporter\Writer\TypedWriterInterface;
+use Sonata\Exporter\Source\SourceIteratorInterface;
+use Sonata\Exporter\Writer\TypedWriterInterface;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 
 /**
@@ -91,3 +91,5 @@ final class Exporter
         $this->writers[$writer->getFormat()] = $writer;
     }
 }
+
+class_exists(\Exporter\Exporter::class);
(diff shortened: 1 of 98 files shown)


src/Exporter.php
```diff
@@ -9,10 +9,10 @@
  * file that was distributed with this source code.
  */
 
-namespace Exporter;
+namespace Sonata\Exporter;
 
-use Exporter\Source\SourceIteratorInterface;
-use Exporter\Writer\TypedWriterInterface;
+use Sonata\Exporter\Source\SourceIteratorInterface;
+use Sonata\Exporter\Writer\TypedWriterInterface;
```
kunicmarko20: you don't have to trigger class exists here?

greg0ire: Good catch! It's the only one, I checked with `grep -L`

kunicmarko20: Are you sure? I think I saw another one, will recheck later, on my phone now

greg0ire: You're right, gonna retry that grep :(

greg0ire: Ok this time I found many more…

andrewmy:
Tested on a couple admin classes in one project, nothing appears to be broken 👍

kunicmarko20:
Thank you @andrewmy!

PR #289, opened 2019-02-21
Delete .styleci.yml

OskarStark:


OskarStark:
> Title ismi

hmm? 🤔

kunicmarko20:
Wanted to say that title is misleading since you fix the cs issues.

Issue #286, opened 2019-01-30
DoctrineORMQuerySourceIterator is now final


mpoiriert:
#### Sonata packages

```
$ composer show --latest 'sonata-project/*'
# Put the result here.

sonata-project/admin-bundle              3.45.1 3.45.1 The missing Symfony Admin Generator
sonata-project/block-bundle              3.14.0 3.14.0 Symfony SonataBlockBundle
sonata-project/cache                     2.0.1  2.0.1  Cache library
sonata-project/core-bundle               3.15.1 3.15.1 Symfony SonataCoreBundle
sonata-project/datagrid-bundle           2.4.0  2.4.0  Symfony SonataDatagridBundle
sonata-project/doctrine-extensions       1.1.5  1.1.5  Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.8.1  3.8.1  Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/easy-extends-bundle       2.5.0  2.5.0  Symfony SonataEasyExtendsBundle
sonata-project/exporter                  2.0.1  2.0.1  Lightweight Exporter library
sonata-project/formatter-bundle          3.5.0  4.1.2  Symfony SonataFormatterBundle
sonata-project/media-bundle              3.18.1 3.18.1 Symfony SonataMediaBundle
sonata-project/user-bundle               4.2.3  4.2.3  Symfony SonataUserBundle

```

#### Symfony packages

```
$ composer show --latest 'symfony/*'
# Put the result here.
symfony/dependency-injection v4.2.2             v4.2.2             Symfony DependencyInjection Component
symfony/doctrine-bridge      v4.2.2             v4.2.2             Symfony Doctrine Bridge
symfony/dom-crawler          v4.2.2             v4.2.2             Symfony DomCrawler Component
symfony/dotenv               v4.2.2             v4.2.2             Registers environment variables from a .env file
symfony/event-dispatcher     v4.2.2             v4.2.2             Symfony EventDispatcher Component
symfony/expression-language  v4.2.2             v4.2.2             Symfony ExpressionLanguage Component
symfony/filesystem           v4.2.2             v4.2.2             Symfony Filesystem Component
symfony/finder               v4.2.2             v4.2.2             Symfony Finder Component
symfony/flex                 v1.1.8             v1.1.8             Composer plugin for Symfony
symfony/form                 v4.2.2             v4.2.2             Symfony Form Component
[...]

PR #291, opened 2019-04-09
remove final class

zearg:
## Allowing extends _Handler_ and _Exporter_ classes

Removing **final** keyword on _Handler_ and _Exporter_ classes

I am targeting this branch, because it's backward compatible.

## Motivation

I know it goes against the meaning of changelog 2.x

In my case, i want to give more data to a customized writer (want to give a docx template, datas, schema and files).
I need to call a newly created **getWriter(string $format)** method on **Exporter**. 

```php
public function getWriter(string $format)
{
    if(isset($this->writers[$format]) && $this->writers[$format] !== null){
        return $this->writers[$format];
    }

    return false;
}
```
Without extending, i have to rewrite the whole **Exporter** class. 
And Decorator pattern doesn't works in this case... (cant access to **private $writers;** from an **ExporterDecorator** class)

This newly created method is used from a service that i create (in symfony), where **Exporter** is instantiated in _$this->exporter_ :

```php
$writer = $this->exporter->getWriter($format);
$writer
    ->setTemplatePath($template)
    ->setSchema($schema)
    ->setFiles($files)
;
```

My writer is a DocxWriter that i created.

I think i've a similar case of https://github.com/sonata-project/exporter/issues/286

## Changelog

Minimal modifications, no changelog.

## To do
    
- [x] Update the tests
- [ ] Add an upgrade note

### DIFF
--- a/src/Exporter.php
+++ b/src/Exporter.php
@@ -20,7 +20,7 @@ use Symfony\Component\HttpFoundation\StreamedResponse;
 /**
  * @author Grégoire Paris <postmaster@greg0ire.fr>
  */
-final class Exporter
+class Exporter
 {
     /**
      * @var TypedWriterInterface[]
--- a/src/Handler.php
+++ b/src/Handler.php
@@ -16,7 +16,7 @@ namespace Sonata\Exporter;
 use Sonata\Exporter\Source\SourceIteratorInterface;
 use Sonata\Exporter\Writer\WriterInterface;
 
-final class Handler
+class Handler
 {
     /**
      * @var SourceIteratorInterface


OskarStark:
Uhh, why want you do this? 🤔

zearg:
Sorry, i edit my first post adding my motivation 

## Motivation

I know it goes against the meaning of changelog 2.x

In my case, i want to give more data to a customized writer (a docx template, datas, schema and files).
I need to call a newly **getWriter(string $format)** method on **Exporter**. 
Without extending, i have to rewrite the whole Exporter class. 
And Decorator pattern doesn't works in this case... (cant access to **private $writers;** from **Exporter** class)



I think i've a similar case of https://github.com/sonata-project/exporter/issues/286

zearg:
Sorry, maybe i don't understand.

I did an ExporterDecorator class with 

`use Sonata\Exporter\Source\SourceIteratorInterface;
use Sonata\Exporter\Writer\TypedWriterInterface;
use Sonata\Exporter\Exporter;

class CustomExporter
{
    private $exporter;

    private $writers;

    public function __construct(Exporter $exporter, $writers) {
        $this->exporter = $exporter;

        foreach ($writers as $writer) {
            $this->exporter->addWriter($writer);
            $this->writers[$writer->getFormat()] = $writer;
        }
    }

    public function addWriter(TypedWriterInterface $writer) {
        $this->exporter->addWriter($writer);
    }

    public function getResponse($format, $filename, SourceIteratorInterface $source) {
        return $this->exporter->getResponse($format, $filename, $source);
    }

    public function getAvailableFormats() {
        return $this->exporter->getAvailableFormats();
    }

    public function getWriter(string $format)
    {
        if(isset($this->writers[$format]) && $this->writers[$format] !== null){
            return $this->writers[$format];
        }

        // todo : exception error
        return false;
    }
}`

Is it the right way to do it ?
Is it easier than an extended class ? :s

core23:
> Is it the right way to do it ?
> Is it easier than an extended class ? :s

Yes, that's the right way. You should ask yourself if you really need all methods for your problem or if you could write an adapter that provides only the methods you need.

greg0ire:
I read the whole thread again and I'm not even sure you need all this.

1. create a service for your writer
2. inject it in the exporter
3. from your piece of code, get the service created in 1 by its name
4. configure it

No need for `getWriter` at all.

zearg:
I think I need to be more specific about my intention to use the sontata exporter project :

At the moment I have 4 writers I use : docx, pdf, xls, csv.
That's why i use Sonata Exporter.
But 2 formats (docx and pdf) used templates for generation.

I created a service where i do 
```php
switch ($format) {
    case self::FORMAT_PDF:
    case self::FORMAT_DOCX:
        /** @var  $writer */
        $writer = $this->exporter->getWriter($format);
        $writer
            ->setTemplatePath($templatePath)
            ->setSchema($schema)
            ->setFiles($files)
        ;
        return $this->getZipResponse($format, $filename, $data) ?? $this->exporter->getResponse($format, $filename, new ArraySourceIterator($data));
        break;
    default:
        return $this->exporter->getResponse($format, $filename, new ArraySourceIterator($data));
        break;
}
```

Create a service for docx **and** pdf will works but make gas plant.
It's for this simple reason that I wanted to be able to extend the Export class

core23:
>  It's for this simple reason that I wanted to be able to extend the Export class

Please specify, which extension point you need for your problem. Do you need a hook, getter or setter?

zearg:
I want to add a getter (**getWriter($format)**) on **Exporter** class, with $format argument.
```php
$exporter = new Exporter();
$exporter->getWriter('docx'); // Return DocxWriter
$exporter->getWriter('pdf'); // Return PdfWriter
$exporter->getWriter('csv'); // Return CsvWriter
....
```

To pass more informations to my writer (docx template or pdf template and a $schema to know if each data has to be converted from markdown to text for example).

Iit's more complex than for csv and xls

greg0ire:
I think your use case is valid if `$templatePath`, `$schema` or `$files` can change in the same deploy (from one php execution to another, at runtime). Is it the case?

zearg:
Absolutely, I create an exporter micro-service that can take a table of data, a template, files (images) and a schema to indicate what are the data.

In every request to export a .pdf or a .docx I will pass $data, $schema, $files and $template to my controller (i use symfony).

greg0ire:
Let's say your docx writer service is called `foo-bar-42`. You could completely inject it in your controller _and_ in the exporter, right? It would be the very same object both times. So you would have this:
```
// use statements not included because I'm to tired for this

final class MyRandomAction
{
    public function __construct(DocxWriter $writer, Exporter $exporter)
    {
         // assignment to local variables
    }

    public function __invoke()
    {
         $this->writer->setTemplate($template)->setAndSoOn($otherRandomVar);

         return $this->getZipResponse($format, $filename, $data) ??
             $this->exporter->getResponse($format, $filename, new ArraySourceIterator($data));
    }
}
```

And in your DI:

```
    My\Action\MyRandomAction:
        $writer: '@foo-bar-42' # use autowiring for the exporter

    foo-bar-42: # that service, that object will be injected in the exporter. It's the very same object that is injected in your controller
        class: MyCompany\MyNamespace\DocxWriter
        tag: {name: sonata.exporter.writer}
```

and that's it. No need for extension, decoration or whatnot.

If that's not good enough, I think you could contribute a `getWriter` method, it's far more likely to be accepted than the removal of that `final` keyword.

greg0ire:
Also, another solution you could use is to pass as a third argument an object implementing `SourceIteratorInterface` *and* another interface:

```
MyInterface
{
    public function templatePath();
    public function schema();
    public function files();
}
```

Here too, no need for extension / decoration, and since you seem to pass the same values regardless of whether it's a pdf or a docs, this seems to make sense.

zearg:
Thanks, yes I can use a custom **SourceIteratorInterface** to pass data (original $data) + custom datas ($template, $schema....).

But finally, i used your method where i inject writers in my service. I will perhaps do your last method when I have time.
The only bad point is that we inject the writers twice. Once in Sonata Export and a second time in the ExportService.

I think adding a **getWriter($format)** method on Exporter will be a good thing. I think i'll make a PR that goes in this direction.

But out of curiosity, being able to expand the class would also have worked very well. Why pass the classes in final?

greg0ire:
> I think adding a getWriter($format) method on Exporter will be a good thing. I think i'll make a PR that goes in this direction.

Not that good: you have no guarantee you will end up with the class you want, unless you know how the object was build, and this breaks encapsulation.


But out of curiosity, being able to expand the class would also have worked very well. Why pass the classes in final?

It makes our life as maintainers much easier, because there are far fewer BC-breaks, which allows us to change the design as we want. Also, it encourages encapsulation. More on this here: https://ocramius.github.io/blog/when-to-declare-classes-final/

zearg:
Okay, thanks for explanations :)

zearg:
I have another problem

I wanted to explain it to you to say that everything would be simpler if we removed this key word, but I will seek a solution to circumvent the problem

Sorry for reopening...

greg0ire:
It's always a bit weird at first but it often leads to more testable, more reusable. We're here to help if you're stuck.

zearg:
I'm facing another problem : 

I'm using PHPUnit, i want to mock Exporter class, but we can't mock final class.

greg0ire:
Don't mock what you don't own, use the real thing, it has no dependencies.

PR #310, opened 2019-12-03
Add throw annotation

VincentLanglet:
## Subject

I am targeting this branch, because it's just phpdoc.

`open` and `write` function can throws \RuntimeException. Let's warn about this.
When using a tool checking for try/catch, a wrong phpdoc give a weird behaviour.

## Changelog

Not needed for phpdoc

### DIFF
--- a/src/Writer/CsvWriter.php
+++ b/src/Writer/CsvWriter.php
@@ -65,6 +65,9 @@ final class CsvWriter implements TypedWriterInterface
      */
     private $terminate;
 
+    /**
+     * @throws \RuntimeException
+     */
     public function __construct(
         string $filename,
         string $delimiter = ',',
@@ -115,6 +118,9 @@ final class CsvWriter implements TypedWriterInterface
         fclose($this->file);
     }
 
+    /**
+     * @throws InvalidDataFormatException
+     */
     public function write(array $data): void
     {
         if (0 === $this->position && $this->showHeaders) {
--- a/src/Writer/CsvWriter.php
+++ b/src/Writer/CsvWriter.php
@@ -118,9 +118,6 @@ final class CsvWriter implements TypedWriterInterface
         fclose($this->file);
     }
 
-    /**
-     * @throws InvalidDataFormatException
-     */
     public function write(array $data): void
     {
         if (0 === $this->position && $this->showHeaders) {
--- a/src/Writer/XmlWriter.php
+++ b/src/Writer/XmlWriter.php
@@ -45,6 +45,9 @@ final class XmlWriter implements TypedWriterInterface
      */
     private $childElement;
 
+    /**
+     * @throws \RuntimeException
+     */
     public function __construct(string $filename, string $mainElement = 'datas', string $childElement = 'data')
     {
         $this->filename = $filename;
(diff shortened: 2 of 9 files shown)


src/Writer/CsvWriter.php
```diff
@@ -65,6 +65,9 @@ final class CsvWriter implements TypedWriterInterface
      */
     private $terminate;
 
+    /**
+     * @throws \RuntimeException
```
core23: Maybe it's better to add this (or a more specific domain exception) to the interface?

VincentLanglet: You mean adding

```
    /**
     * @throws \RuntimeException
     */
    public function __construct();
```
?

But how can I do this since the number of arguments is not the same ?

core23: A `RuntimeException` should be okay for the constructor, but the other methods (`open`, `write`, ..) should throw a `WriterException`

VincentLanglet: The PR is updated

src/Writer/XmlWriter.php
```diff
@@ -80,6 +84,9 @@ public function close(): void
         fclose($this->file);
     }
 
+    /**
+     * @throws SonataExporterException
```
core23: You don't need this, when you add this to the interface

core23: You should also wrap the existing `RuntimeException` inside the `SonataExporterException`

VincentLanglet: What do you mean by the second message ?

VincentLanglet: Ok, I create a generic `class RuntimeException extends \RuntimeException implements SonataExporterException`.

This way when creating a custom writer it's easier to do
```
    public function open()
    {
        try {
            $this->myWriterFromExtLibrary->openButCanThrowException();
        } catch (\Exception $e) {
            throw new Sonata\Exporter\Exception\RuntimeException()
        }
    }
```
without having to create a custom SonataExporterException too

VincentLanglet:
@core23 I have reread this PR, and I'm wondering why did you prefer throw a `SonataExporterException` instead of a simple `Exception`.

This enforce
```
/**
 * @throws SonataExporterException
 **/
public function open()
    {
        try {
            $this->myWriterFromExtLibrary->openButCanThrowException();
        } catch (\Exception $e) {
            throw new Sonata\Exporter\Exception\RuntimeException($e->getMessage())
        }
    }
```
Instead of
```
/**
 * @throws Exception
 **/
public function open()
    {
        $this->myWriterFromExtLibrary->openButCanThrowException();
    }
```

OskarStark:
Because it’s part of our „domain“

core23:
> Because it’s part of our „domain“

That's exactly the reason. A domain exception is easier to catch and could occur in specific cases. Otherwise you could not distinguish between a real `RuntimeException` and a `RuntimeException` we expect.

VincentLanglet:
Ok ! I let the `SonataExporterException` then.
Does the PR need changes ?

VincentLanglet:
Hi @greg0ire. Do you have some time in the following day to make a new release with this PR ? Thanks ! :)

greg0ire:
I'm on it, buddy!

> Not needed for phpdoc

It's a minor, there should be a changelog. I will try to invent one.

VincentLanglet:
@greg0ire Oh sorry, when I started this PR, it was a patch for me.

```
XmlWriter not throw a Sonata RuntimeException instead of the classic RuntimeException
```

greg0ire:
No problem, I made something up :P

PR #341, opened 2020-06-11
Introduce Export field

VincentLanglet:
## Subject

I am targeting this branch, because {reason}.

Closes #{put_issue_number_here}.

## Changelog

```markdown
### Added
- Added some `Class::newMethod()` to do great stuff.

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

### DIFF
--- a/src/Field/ExportFieldInterface.php
+++ b/src/Field/ExportFieldInterface.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\Exporter\Field;
+
+use Sonata\Exporter\Exception\InvalidArgumentException;
+
+interface ExportFieldInterface
+{
+    public function __construct(string $label, string $path, array $options = []);
+
+    public function getLabel(): string;
+
+    public function getPath(): string;
+
+    public function getOptions(): array;
+
+    /**
+     * @param mixed $value
+     *
+     * @throws InvalidArgumentException when the $value type is not supported
+     */
+    public function formatValue($value): ?string;
+}
--- a/src/Source/AbstractPropertySourceIterator.php
+++ b/src/Source/AbstractPropertySourceIterator.php
@@ -1,0 +1,191 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\Exporter\Source;
+
+use Doctrine\ORM\Internal\Hydration\IterableResult;
+use Doctrine\ORM\Query;
+use Sonata\Exporter\Field\ExportFieldInterface;
+use Symfony\Component\PropertyAccess\Exception\UnexpectedTypeException;
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\PropertyAccess\PropertyAccessor;
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+/**
+ * Class DoctrineORMQuerySourceIterator
+ *
+ * @final
+ */
+abstract class AbstractPropertySourceIterator implements SourceIteratorInterface
+{
+    private const DATE_PARTS = [
+        'y' => 'Y',
+        'm' => 'M',
+        'd' => 'D',
+    ];
+    private const TIME_PARTS = [
+        'h' => 'H',
+        'i' => 'M',
+        's' => 'S',
+    ];
+
+    /**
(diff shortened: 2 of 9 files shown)


src/Source/AbstractPropertySourceIterator.php
```diff
@@ -0,0 +1,207 @@
...
+            case \is_object($value):
+                return method_exists($value, '__toString') ? (string) $value : null;
+            case \is_bool($value):
+                // TODO: trans
+
+                return $value;
+            case \is_string($value):
+                // TODO: translate
+
+                return $value;
```
VincentLanglet, line 174: Hi @sonata-project/contributors.

I need some help/advice.
Is there a way to use the symfony translator here ?

wbloszyk: IMHO you can use `$option['translation_domain']` or other `add` method.

VincentLanglet: Yes, i wanna try to introduce an ExportMapper.

```
public function configureExportFields(ExportMapper $exportMapper)
    {
        $exportMapper->add('foo', CustomListType::class, [
            'custom_param' => ['param1' => 'test' ]
        ]);
        $exportMapper->add('bar', 'trans');
        $exportMapper->add('baz', null, ['label' => 'special_label']);
    }
```

With 
```
public add($field, $type, $options)
{
   $label = $options['label'] ?? $field;
   $export = new ExportField($label, $field, $options)

   // ...
}
```

I gonna update the PR/notes, when the PR is ready.

wbloszyk: You can add type classes and there translate. Something simmilar to form field.

VincentLanglet: > If yes you can check how list is translate.

My issue is that for the show/list/..., the translator is coming from the twig filter `|trans`, or sometimes from the `$translator` property in the Admin.

Here, in this class, I currently don't have access to a translator property. I was looking for a way to solve this.
The current usage of the QuerySourceIterator is here 
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Model/ModelManager.php#L478
called here
https://github.com/sonata-project/SonataAdminBundle/blob/a0f815bb2a51cdf041956fb595d027b45dcba4c5/src/Admin/AbstractAdmin.php#L669

The current solution I found would be to add methods like
```
setTranslator
setTranslationDomain
```
to the SourceIterator and call them in the AbstractAdmin, when creating the SourceIterator.

And then I could do something like
```
if ($option['trans'] = true) {
    if (null === $this->translator) {
        throw \Exception()
    } else {
        $value = $this->translator->trans($value);
    }
}
```

But I'm not a big fan of this solution, so I was wondering if there was a better way to use the translator in the SourceIterator. @wbloszyk

wbloszyk: Mybe you dont have to translate it on getValue methods. It will be even better do it while user call export method. For example user will be able generate diffrence  language version based on one time loaded data.

src/Source/AbstractPropertySourceIterator.php
```diff
@@ -0,0 +1,191 @@
...
+use Doctrine\ORM\Internal\Hydration\IterableResult;
+use Doctrine\ORM\Query;
+use Sonata\Exporter\Field\ExportFieldInterface;
+use Symfony\Component\PropertyAccess\Exception\UnexpectedTypeException;
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\PropertyAccess\PropertyAccessor;
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+/**
+ * Class DoctrineORMQuerySourceIterator
```
core23, line 25: This is a useless comment (and wrong too)

src/Source/AbstractPropertySourceIterator.php
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * @final
```
core23, line 27: final and abstract?

src/Source/AbstractPropertySourceIterator.php
```diff
@@ -0,0 +1,191 @@
...
+    private const TIME_PARTS = [
+        'h' => 'H',
+        'i' => 'M',
+        's' => 'S',
+    ];
+
+    /**
+     * @var Query
+     */
+    protected $query;
```
core23, line 45: Please make this private by default

VincentLanglet: $query is not used here, I should remove it. Others field can indeed be private

src/Source/AbstractPropertySourceIterator.php
VincentLanglet: Bad copy paste from DoctrineORMQuerySourceIterator

src/Field/ExportFieldInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\Exporter\Field;
+
+use Sonata\Exporter\Exception\InvalidArgumentException;
+
+interface ExportFieldInterface
+{
+    public function __construct(string $label, string $path, array $options = []);
```
core23, line 20: The constructor should never be a part of an interface

core23 (changes requested):

PR #343, opened 2020-07-10
Add support for array and traversable

### LINKED ISSUE #31: Handle type array in the DoctrineORMQuerySourceIterator
Hi,

i'm playing with Sonata Admin Bundle, the export feature use this library to output a file  in different formats : json,csv . I saw that "array type" for a doctrine entity is handled by returning $value=null. ([line105 DoctrineORMQuerySourceIterator.php](https://github.com/sonata-project/exporter/blob/master/lib/Exporter/Source/DoctrineORMQuerySourceIterator.php#L105)). So the json exported of an entity that have fields/field of "type array" is not complete. I want to ask why we can't return the array?

VincentLanglet:
## Subject

I am targeting this branch, because BC.

Closes #31.

## Changelog

```markdown
### Added
- Added support for array and traversable in `DoctrineORMQuerySourceIterator`, `DoctrineODMQuerySourceIterator` and `PropelCollectionSourceIterator`.
```

### DIFF
--- a/src/Source/AbstractPropertySourceIterator.php
+++ b/src/Source/AbstractPropertySourceIterator.php
@@ -48,12 +48,12 @@ abstract class AbstractPropertySourceIterator implements SourceIteratorInterface
     protected $dateTimeFormat;
 
     /**
-     * @var array<string>
+     * @var string[]
      */
     protected $fields = [];
 
     /**
-     * @param array<string> $fields Fields to export
+     * @param string[] $fields Fields to export
      */
     public function __construct(array $fields, string $dateTimeFormat = 'r')
     {
@@ -126,7 +126,7 @@ abstract class AbstractPropertySourceIterator implements SourceIteratorInterface
     }
 
     /**
-     * @param object|array $current
+     * @param object|mixed[] $current
      */
     protected function getCurrentData($current): array
     {
@@ -161,8 +161,9 @@ abstract class AbstractPropertySourceIterator implements SourceIteratorInterface
     {
         switch (true) {
             case \is_array($value):
+                return '['.implode(', ', array_map([$this, 'getValue'], $value)).']';
             case $value instanceof \Traversable:
-                return null;
+                return '['.implode(', ', array_map([$this, 'getValue'], iterator_to_array($value))).']';
             case $value instanceof \DateTimeInterface:
                 return $value->format($this->dateTimeFormat);
             case $value instanceof \DateInterval:
--- a/src/Source/AbstractPropertySourceIterator.php
+++ b/src/Source/AbstractPropertySourceIterator.php
@@ -1,0 +1,176 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\Exporter\Source;
+
+use Doctrine\ORM\Internal\Hydration\IterableResult;
+use Symfony\Component\PropertyAccess\Exception\UnexpectedTypeException;
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\PropertyAccess\PropertyAccessor;
+use Symfony\Component\PropertyAccess\PropertyPath;
+
+abstract class AbstractPropertySourceIterator implements SourceIteratorInterface
+{
+    private const DATE_PARTS = [
+        'y' => 'Y',
+        'm' => 'M',
+        'd' => 'D',
+    ];
+    private const TIME_PARTS = [
+        'h' => 'H',
+        'i' => 'M',
+        's' => 'S',
+    ];
+
+    /**
+     * @var IterableResult
+     */
+    protected $iterator;
+
+    /**
+     * @var PropertyAccessor
+     */
+    protected $propertyAccessor;
+
+    /**
+     * @var string default DateTime format
+     */
+    protected $dateTimeFormat;
+
+    /**
+     * @var array<string>
+     */
+    protected $fields = [];
+
+    /**
+     * @param array<string> $fields Fields to export
+     */
+    public function __construct(array $fields, string $dateTimeFormat = 'r')
+    {
+        $this->propertyAccessor = PropertyAccess::createPropertyAccessor();
+        $this->fields = $fields;
+
+        $this->dateTimeFormat = $dateTimeFormat;
+    }
+
+    public function current()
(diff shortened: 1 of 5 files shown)


src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -13,55 +13,21 @@
...
 use Doctrine\ORM\Query;
 use Sonata\Exporter\Exception\InvalidMethodCallException;
-use Symfony\Component\PropertyAccess\Exception\UnexpectedTypeException;
-use Symfony\Component\PropertyAccess\PropertyAccess;
-use Symfony\Component\PropertyAccess\PropertyAccessor;
-use Symfony\Component\PropertyAccess\PropertyPath;
 
-class DoctrineORMQuerySourceIterator implements SourceIteratorInterface
+/**
+ * @final since sonata/exporter 2.x
```
phansys, line 20:
```suggestion
 * @final since sonata-project/exporter 2.x.
```

tests/Source/DoctrineORMQuerySourceIteratorTest.php
```diff
@@ -52,8 +52,8 @@ public function getValueProvider()
         $dateTimeImmutable = new \DateTimeImmutable();
 
         $data = [
-            [[1, 2, 3], null],
-            [new \ArrayIterator([1, 2, 3]), null],
+            [[1, 2, 3], '[1, 2, 3]'],
+            [new \ArrayIterator([1, 2, 3]), '[1, 2, 3]'],
```
phansys, line 56:
```suggestion
            [new \ArrayIterator([1, 2, 3]), '[1, 2, 3]'],
            [(static function () { yield from [1, 2, 3]; })(), '[1, 2, 3]'],
```

tests/Source/DoctrineORMQuerySourceIteratorTest.php
```diff
@@ -52,8 +52,8 @@ public function getValueProvider()
         $dateTimeImmutable = new \DateTimeImmutable();
 
         $data = [
-            [[1, 2, 3], null],
-            [new \ArrayIterator([1, 2, 3]), null],
```
phansys, line 56: Why did you remove this case?

VincentLanglet: Oh, miss read your comment.

src/Source/AbstractPropertySourceIterator.php
```diff
@@ -0,0 +1,177 @@
...
+    protected function getCurrentData($current): array
+    {
+        $data = [];
+        foreach ($this->fields as $key => $field) {
+            if (\is_string($field)) {
+                $name = \is_string($key) ? $key : $field;
+                $propertyPath = $field;
+            } else {
+                throw new \TypeError('Unsupported field type. Field should be a string.');
+            }
```
phansys, line 140:
```suggestion
            if (!\is_string($field)) {
                throw new \TypeError(sprintf(
                    'Field type must be string, %s given.',
                    \is_object($field) ? 'instance of '.\get_class($field) : \gettype($field)
                );
            }

            $name = \is_string($key) ? $key : $field;
            $propertyPath = $field;
```

VincentLanglet: I plan to support another type https://github.com/sonata-project/exporter/pull/341

phansys: Great. IMO, if we move the exception before there is no need for the `else` clause. When the new type gets support, an `or` condition could be added (`if (!\is_string($field) && !$field instanceof ExportFieldInterface)`).

VincentLanglet: This will be
```
if (\is_string($field) {
    doThings();
else ($field instanceof ExportFieldInterface) {
    doOtherThings();
} else {
    throw new Exception();
}
```

src/Source/AbstractPropertySourceIterator.php
```diff
@@ -0,0 +1,177 @@
...
+     */
+    protected $propertyAccessor;
+
+    /**
+     * @var string default DateTime format
+     */
+    protected $dateTimeFormat;
+
+    /**
+     * @var array<string>
```
greg0ire, line 51: IMO we should either use

```suggestion
     * @var string[]
     * @psalm-var list<string>
```

or 

```suggestion
     * @var list<string>
```

or

```suggestion
     * @var string[]
```

src/Source/AbstractPropertySourceIterator.php
```diff
@@ -0,0 +1,177 @@
...
+
+        if ('' === $datePart && '' === $timePart) {
+            return 'P0Y';
+        }
+
+        return 'P'.$datePart.('' !== $timePart ? 'T'.$timePart : '');
+    }
+
+    /**
+     * @param object|array $current
```
greg0ire, line 129: array of what? `mixed`?

src/Source/AbstractPropertySourceIterator.php
```diff
@@ -0,0 +1,177 @@
...
+            }
+        }
+
+        return $data;
+    }
+
+    /**
+     * @param mixed $value
+     *
+     * @return string|null
```
greg0ire, line 158: Why is this not a return type hint?

src/Source/AbstractPropertySourceIterator.php
VincentLanglet: I don't know, so yeah mixed maybe. I took the typehint from property accessor getValue.

src/Source/AbstractPropertySourceIterator.php
VincentLanglet: To be BC. The method was already existing.

src/Source/AbstractPropertySourceIterator.php
VincentLanglet: I wanted to introduce an ExportField type, so I would change to `@var array<string|ExportField>`.
But I can use `string[]` atm if you prefer.

VincentLanglet:
Phpdoc is updated @greg0ire :)

src/Source/AbstractPropertySourceIterator.php
```diff
@@ -0,0 +1,177 @@
...
+
+        if ('' === $datePart && '' === $timePart) {
+            return 'P0Y';
+        }
+
+        return 'P'.$datePart.('' !== $timePart ? 'T'.$timePart : '');
+    }
+
+    /**
+     * @param object|array<mixed> $current
```
greg0ire, line 129:
```suggestion
     * @param object|mixed[] $current
```

VincentLanglet: Oops!...I Did It Again

Fixed !

PR #371, opened 2020-09-18
Improve rewind

VincentLanglet:
## Subject

I am targeting this branch, because BC.

I allowed to use multiple times the `rewind` function for most of our SourceIterator.

Doctrine doesn't allow to call multiple times rewing, but since we're creating a new iterator, we just call rewing once on each iterator.

## Changelog

```markdown
### Added
- Add the ability to call the `rewind()` method multiple times for `DoctrineDBALConnectionSourceIterator`, `DoctrineODMQuerySourceIterator`, `DoctrineORMQuerySourceIterator`.
```

### DIFF
--- a/src/Source/AbstractPropertySourceIterator.php
+++ b/src/Source/AbstractPropertySourceIterator.php
@@ -33,7 +33,7 @@ abstract class AbstractPropertySourceIterator implements SourceIteratorInterface
     ];
 
     /**
-     * @var IterableResult
+     * @var IterableResult|null
      */
     protected $iterator;
 
--- a/src/Source/DoctrineORMQuerySourceIterator.php
+++ b/src/Source/DoctrineORMQuerySourceIterator.php
@@ -14,7 +14,6 @@ declare(strict_types=1);
 namespace Sonata\Exporter\Source;
 
 use Doctrine\ORM\Query;
-use Sonata\Exporter\Exception\InvalidMethodCallException;
 
 /**
  * @final since sonata-project/exporter 2.x.
@@ -49,10 +48,6 @@ class DoctrineORMQuerySourceIterator extends AbstractPropertySourceIterator impl
 
     final public function rewind(): void
     {
-        if ($this->iterator) {
-            throw new InvalidMethodCallException('Cannot rewind a Doctrine\ORM\Query');
-        }
-
         $this->iterator = $this->query->iterate();
         $this->iterator->rewind();
     }
--- a/src/Source/PropelCollectionSourceIterator.php
+++ b/src/Source/PropelCollectionSourceIterator.php
@@ -39,13 +39,10 @@ final class PropelCollectionSourceIterator extends AbstractPropertySourceIterato
 
     public function rewind(): void
     {
-        if ($this->iterator) {
-            $this->iterator->rewind();
-
-            return;
+        if (!$this->iterator) {
+            $this->iterator = $this->collection->getIterator();
         }
 
-        $this->iterator = $this->collection->getIterator();
         $this->iterator->rewind();
     }
 }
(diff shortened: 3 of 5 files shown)


greg0ire:
Please improve your commit message.

greg0ire:
Sorry didn't mean to click "close" :sweat_smile:

VincentLanglet:
@greg0ire I changed the commit message

src/Source/AbstractPropertySourceIterator.php
```diff
@@ -33,7 +33,7 @@ abstract class AbstractPropertySourceIterator implements SourceIteratorInterface
     ];
 
     /**
-     * @var IterableResult
+     * @var IterableResult|null
```
greg0ire, line 36: How does this relate to `rewind()` / this commit?

src/Source/PropelCollectionSourceIterator.php
```diff
@@ -39,13 +39,10 @@ public function __construct(PropelCollection $collection, array $fields, string
...
     {
-        if ($this->iterator) {
-            $this->iterator->rewind();
-
-            return;
+        if (!$this->iterator) {
+            $this->iterator = $this->collection->getIterator();
         }
 
-        $this->iterator = $this->collection->getIterator();
```
greg0ire, line 48: No exception was thrown previously here… would it be fair to say that you could already call `rewind()` multiple times here? If yes, how does what are you doing here exactly, and how does it relate to this commit?

VincentLanglet:
And now you have 3 commits @greg0ire :)

src/Source/AbstractPropertySourceIterator.php
greg0ire: Why is that wrong? Looking at `DoctrineDBALConnectionSourceIterator::__construct()`, it seems to me that this property can be null.

greg0ire (commented):
In your second commit:

consecutiv => consecutive
rewing => rewind

src/Source/AbstractPropertySourceIterator.php
VincentLanglet: Yes, it can be null. That's why I allow the null value.

greg0ire: Oh sorry I read this the wrong way!

greg0ire (approved):

greg0ire (approved):
I pushed commits with explanations for each one, please review in case I misunderstood.

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -49,10 +48,6 @@ public function current()
 
     final public function rewind(): void
     {
-        if ($this->iterator) {
-            throw new InvalidMethodCallException('Cannot rewind a Doctrine\ORM\Query');
-        }
-
         $this->iterator = $this->query->iterate();
```
franmomu, line 51: Now I see this, [`Query::iterate` was deprecated in `doctrine/mongo-odm` `1.3`](https://github.com/doctrine/mongodb-odm/blob/1.3.x/lib/Doctrine/ODM/MongoDB/Query/Query.php#L413), we should add a constraint and change the method (in another PR). 

AFAIK, in the ODM returns a [Cursor which is rewindable](https://github.com/doctrine/mongodb/blob/master/lib/Doctrine/MongoDB/Cursor.php#L529).

VincentLanglet: Is this PR ok for you then ?

You made your comment on `DoctrineORMQuerySourceIterator`, but the deprecation is on ODM.

Do you want to create the PR with the migration from iterate to getIterator ?
We also have 0 test about `ODM`, maybe it could be great to add some.

franmomu (approved):

franmomu:
Thanks @VincentLanglet !

PR #373, opened 2020-09-20
Add keywords

OskarStark:


VincentLanglet:
We should have some consistency @OskarStark.

Here you're using `bundle` and `symfony-bundle`.
In sonataAdmin, I use `bundle` and `symfony`, which are more popular.

PR #394, opened 2020-10-12
Test parent class directly

VincentLanglet:
Although DoctrineORMQuerySourceIteratorTest is named after
DoctrineORMQuerySourceIterator, it only tests getValue(), which is
inherited from AbstractPropertySourceIterator.
We can easily test that method without using
DoctrineORMQuerySourceIterator, but we still need to extend that class
because getValue() is protected, which means we have to make it public
to test it.
As a bonus, we stop extending the class DoctrineORMQuerySourceIterator,
which will be final in next major.
This allows us to avoid a warning from PHPUnit.

Fix https://github.com/sonata-project/exporter/pull/393#issuecomment-706985719


VincentLanglet:
BTW @jordisala1991 Test are failing on PHP 8

But mongodb seems installed.
```
- name: Install PHP with extensions
        uses: shivammathur/setup-php@v2
        with:
          php-version: ${{ matrix.php-version }}
          coverage: pcov
          tools: composer:v2, pecl
          extensions: mongodb
```
Is there something to do ? Is it related to the following hack ?
```
- name: Configuration required for PHP 8.0
        if: matrix.php-version == '8.0'
        run: composer config platform.php 7.4.99 && composer require phpunit/phpunit:"9.3.*" --no-update
```

greg0ire:
I've been working on migrating doctrine/dbal to GA recently, and I noticed that the PHP action will not fail even when it fails to install an extension. not sure why.

> But mongodb seems installed.

Nope:


> ==> Setup Extensions
> ✗ mongodb Could not install mongodb on PHP 8.0.0-dev

jordisala1991:
Looks like we are losing coverage? Without the codecov comment it not that easy to see

greg0ire:
Indeed, it's weird to me that there is a `+` there

VincentLanglet:
Certainly related to the fact that `DoctrineORMQuerySourceIterator::construct` is not called anymore.

greg0ire:
> Indeed, it's weird to me that there is a + there

Not weird in fact, coverage may increase in some files, decrease in some others.

greg0ire:
> Is this PR stil ok for you @jordisala1991 @greg0ire ?

I have to click 2 links to find the warning in the test, which is the why of your PR. The commit message does not mention _what_ you are doing to avoid that warning.

VincentLanglet:
@greg0ire I rename the commit

tests/Source/AbstractPropertySourceIteratorTest.php
```diff
@@ -13,36 +13,31 @@
...
+    public function getValue($value)
     {
-        $entityManager = $this->createMock(EntityManager::class);
-        $entityManager->method('getConfiguration')->willReturn(new Configuration());
-        $this->query = new Query($entityManager);
+        return parent::getValue($value);
     }
+}
 
+final class AbstractPropertySourceIteratorTest extends TestCase
```
core23, line 33: Is there a reason why you define two classes in this file?

greg0ire:
> @greg0ire I rename the commit

The what is here now, but you lost the why, and it would be great to have the name of the class to understand further

> Stop extending a final class in tests

This warrants an explanation. I mean, final classes cannot be extended by definition, can they?

greg0ire:
It is better!
Link for the lazy: https://github.com/sonata-project/exporter/blob/dddc8925891eeb295258c8ad8df02789fc99d27c/src/Source/DoctrineORMQuerySourceIterator.php#L18-L21

greg0ire:
It looks like the goal of this test was not to test `DoctrineORMQuerySourceIterator` but its parent class, is that correct?

tests/Source/AbstractPropertySourceIteratorTest.php
greg0ire: Have you considered using an anonymous class for that purpose?

greg0ire:
I think I understood what this is about and pushed an explanation.

VincentLanglet:
Whoah, what a commit message. Thanks @greg0ire 

RTM @sonata-project/contributors

greg0ire:
> RTM

There still is a pending discussion here: https://github.com/sonata-project/exporter/pull/394#discussion_r504195746

tests/Source/AbstractPropertySourceIteratorTest.php
VincentLanglet: I never used an anonymous class. But I don't think it would be better with it. Maybe is it because I'm not used to.

greg0ire: I like to use them for that kind of thing partly because it nips these sort of discussions about several classes per file in the bud but as you wish, I'm fine with both.

greg0ire (approved):

tests/Source/AbstractPropertySourceIteratorTest.php
VincentLanglet: I'm fine with everything. I can change if @sonata-project/contributors prefer the anonymous class, but a gist with the anonymous class could help me a lot.

tests/Source/AbstractPropertySourceIteratorTest.php
```diff
@@ -13,36 +13,31 @@
...
+{
     /**
      * @dataProvider getValueProvider
      */
     public function testGetValue($value, $expected, $dateFormat = 'r'): void
     {
-        $iterator = new DoctrineORMQuerySourceIterator($this->query, [], $dateFormat);
+        $iterator = new PropertySourceIterator([], $dateFormat);
```
greg0ire, line 40: Here is how it would look like
```suggestion
        $iterator = new class([], $dateFormat) extends AbstractPropertySourceIterator
        {
            public function rewind(): void
            {
                $this->iterator->rewind();
            }
            
            public function getValue($value)
            {
                return parent::getValue($value);
            }
        };
```

VincentLanglet: Thanks done

greg0ire (approved):

phansys (approved):

PR #403, opened 2020-11-14
Fix php 8 tests

jordisala1991:


src/Writer/CsvWriter.php
```diff
@@ -126,6 +126,12 @@ public function write(array $data): void
             ++$this->position;
         }
 
+        if (1 !== \strlen($this->delimiter) ||
+            1 !== \strlen($this->enclosure) ||
+            1 !== \strlen($this->escape)) {
+            throw new InvalidDataFormatException();
+        }
```
jordisala1991, line 133: The fputcsv only accepts one character for `delimiter` `enclosure` and `escape`.

Previous php versions would just accept more and return null (not sure if due to the supress operator or always), but php8 throws an exception instead.

To avoid behavioral bc breaks, I just add this check before calling the function. (we could could consider not allowing to do a new `CsvWriter` with more than one char on next major).

tests/ExporterTest.php
```diff
@@ -36,12 +36,8 @@ public function testFilter(): void
 
     public function testConstructorRejectsNonTypedWriters(): void
     {
-        $this->expectException(
-            version_compare(PHP_VERSION, '7.0.0', '<') ? 'PHPUnit_Framework_Error' : 'TypeError'
-        );
-        $this->expectExceptionMessage(
-            'must implement interface'
```
jordisala1991, line 43: The message changes from php 7 to 8. Since this message does not include any sensitive information, IMO it is better to just remove the check for the message and leave only the type of the exception.

tests/ExporterTest.php
```diff
@@ -36,12 +36,8 @@ public function testFilter(): void
 
     public function testConstructorRejectsNonTypedWriters(): void
     {
-        $this->expectException(
-            version_compare(PHP_VERSION, '7.0.0', '<') ? 'PHPUnit_Framework_Error' : 'TypeError'
```
jordisala1991, line 40: We do not support php < 7 anymore..

VincentLanglet (dismissed):

src/Writer/CsvWriter.php
OskarStark: Can you please make this info part of the commit message? Thanks

src/Writer/CsvWriter.php
```diff
@@ -126,6 +126,12 @@ public function write(array $data): void
             ++$this->position;
         }
 
+        if (1 !== \strlen($this->delimiter) ||
+            1 !== \strlen($this->enclosure) ||
+            1 !== \strlen($this->escape)) {
+            throw new InvalidDataFormatException();
```
greg0ire, line 132: Please always provide a message when throwing.
Here is a suggestion:
```suggestion
            throw new InvalidDataFormatException(<<<'EXCEPTION'
            Context: trying to write CSV data
            Problem: delimiter, enclosure or escape character is actually a
            string longer than just one character
            Solution: pick an actual character
            EXCEPTION
            );
```

Even then, it could be improved by showing exactly which character is not just a character.

VincentLanglet: Does this works for php 7.3 ?

greg0ire: [Yes](https://3v4l.org/VG0k0)

OskarStark:
CS is failing, otherwise 👌🏻

src/Writer/CsvWriter.php
```diff
@@ -126,6 +126,18 @@ public function write(array $data): void
             ++$this->position;
         }
 
+        if (1 !== \strlen($this->delimiter) ||
+            1 !== \strlen($this->enclosure) ||
+            1 !== \strlen($this->escape)) {
+            throw new InvalidDataFormatException(<<<'EXCEPTION'
```
greg0ire, line 132:
```suggestion
            throw new InvalidDataFormatException(
            <<<'EXCEPTION'
```

src/Writer/CsvWriter.php
```diff
@@ -126,6 +126,18 @@ public function write(array $data): void
...
+            Context: trying to write CSV data
+            Problem: delimiter, enclosure or escape character is actually a
+            string longer than just one character
+            Solution: pick an actual character
+            EXCEPTION
+            );
```
greg0ire, line 138:
```suggestion
            EXCEPTION);
```

greg0ire (approved):

PR #430, opened 2021-01-06
Prevent csv/xls injection (#429)

### LINKED ISSUE #429: Prevent CSV and XLS injection in Writers
## Feature Request

The export functionality of this bundle allows users to download data in a .csv or .xls file suitable for handling in spreadsheet applications like Microsoft Excel and OpenOffice Calc. The resulting spreadsheet's cells often contain input from untrusted sources such as survey responses, transaction details, and user-supplied content. This is inherently risky, because any cells starting with certain special characters will be interpreted by the spreadsheet software as formulas which can be used to execute payloads. These payloads can compromise a user's system by downloading and running malware as well as giving command line access to a compromised system.

If exported data cells were to contain payloads such as `=cmd|' /C calc'!A0` in a CSV or XLS file generated by the this bundle, this payload becomes embedded within that file and will attempt to run upon being opened (mainly on Windows systems). The payload in this example will open the Windows calculator should a user open the file and accept two prompts (unfortunately many users would do this without questioning it).

Attempting to sanitize data for this specific purpose upon input would be cumbersome as it may come from various input sources. I suggest doing a simple regex filter within the Writer methods for both file formats. I currently have this implemented in another project that is using `sonata-project/exporter:^1.11` by overriding the `write(array $data)` method in each Writer class - however in  `sonata-project/exporter:^2.0` these Writer classes are now marked as final - so I figure the best solution is to propose this as a security feature update to the Sonata Exporter bundle so we can all benefit from it.

timwentzell:
## Subject
Suggested security update to address #429 

This is a simple PR to filter data cells exported to CSV and XLS that start with dangerous characters that can be interpreted by spreadsheet applications.

I am targeting this branch, because each Writer class is marked as final and the easiest solution is to offer this security update to everyone.

Closes #429.

## Changelog

```markdown
### Security
- Prevent csv/xls injection in Writers (#429)
```


greg0ire:
Isn't escaping an option? If I open excel and I want to write a formula verbatim, without it being executed, surely there must be a way to do this?

VincentLanglet:
If I understand you're replacing "-" by "!-" ?

This would break a lot of code IMHO

timwentzell:
> Isn't escaping an option? If I open excel and I want to write a formula verbatim, without it being executed, surely there must be a way to do this?

It seems when editing spreadsheets in Excel directly the general "rule-of-thumb" to prevent a cell from attempting to be executed as a formula is to prepend the cell with an apostrophe. I can update my PR to replace the **!** with an **'** - apparently Excel will hide the single apostrophe, however I can't confirm that since I don't have access to Excel at the moment.

src/Writer/CsvWriter.php
```diff
@@ -137,6 +137,13 @@ public function write(array $data): void
             EXCEPTION);
         }
 
+        // prevent csv injection
+        $patterns = ['/^=/', '/^\+/', '/^-/', '/^@/'];
+        $replace = ['!=', '!+', '!-', '!@'];
+        foreach ($data as $key => $value) {
+            $data[$key] = preg_replace($patterns, $replace, $value);
+        }
```
greg0ire, line 145:
```suggestion
        // prevent csv injection
        foreach ($data as $key => $value) {
            $data[$key] = preg_replace(
                ['/^=/', '/^\+/', '/^-/', '/^@/'];
                ['!=', '!+', '!-', '!@'];
                $value
            );
        }
```

src/Writer/XlsWriter.php
```diff
@@ -78,8 +78,11 @@ public function write(array $data): void
         $this->init($data);
 
         fwrite($this->file, '<tr>');
+        // prevent xls injection
+        $patterns = ['/^=/', '/^\+/', '/^-/', '/^@/'];
+        $replace = ['!=', '!+', '!-', '!@'];
         foreach ($data as $value) {
-            fwrite($this->file, sprintf('<td>%s</td>', $value));
+            fwrite($this->file, sprintf('<td>%s</td>', preg_replace($patterns, $replace, $value)));
```
greg0ire, line 85: Please apply the same here.

timwentzell: Yes good point - no need of those variables being set.

greg0ire:
@timwentzell sorry, I failed to notice that you were already doing some kind of escaping.

@VincentLanglet he's only doing that when the dash is at the start of the line.

src/Writer/CsvWriter.php
greg0ire, line 145: As pointed out by @VincentLanglet , that can break a lot of things. Here, we don't know that this is going to go through Excel, I don't think this writer should be changed.

src/Writer/XlsWriter.php
greg0ire, line 85: In case people actually rely on the values not being escaped, should their be an option to disable the escaping?

timwentzell: Yes I was just thinking maybe this could be a configurable option. I had a client run a pen test on an application using Sonata Admin with the Exporter bundle and they identified this vulnerability - and I had to fix it or else they weren't going to use our application. I didn't like the idea of manipulating the cell data but it was the only option I could find to "plug the hole".

If this isn't useful for the community that's totally fine, for now I've just made my own CsvWriter and XlsWriter classes and use those in place of the ones in this bundle, but that makes me have to closely monitor the changes in those classes when I want to update this bundle - which is no fun at all...

timwentzell: > In case people actually rely on the values not being escaped, should their be an option to disable the escaping?

I have now made it optional, defaulting to not used (**false**) so any existing applications using this bundle will not experience any change unless they explicitly add the **safe_cells** configuration parameter and set it to **true** for one or both writers.

greg0ire:
Instead of adding that option to each and every future writer that produces files that can be opened with Excel, how about we do this (which you can already do in your project, and allows you not to monitor every change, BTW):

```
SafeCellWriterDecorator implements TypedWriterInterface
{
    /** @var TypedWriterInterface */
    private $wrapped;

    public function __construct(TypedWriterInterface $wrapped)
    {
        $this->wrapped = $wrapped;
    }

    public function getDefaultMimeType(): string
    {
        return $this->wrapped->getDefaultMimeType();
    }

    public function getFormat(): string
    {
        return $this->wrapped->getFormat();
    }

    public function open()
    {
        $this->wrapped->open();
    }

    public function close()
    {
        return $this->wrapped->close();
    }

    public function write(array $data)
    {
        // Here, you do your preg_replace()
        return $this->wrapped->write($data);
    }
}
```

Then we use the configuration nodes you introduce to decide whether we use [service decoration](https://symfony.com/doc/current/service_container/service_decoration.html) or not. It will make it easier for you to write tests, by the way.

VincentLanglet (approved):
Seems a lot better !

I only find weird having a csv with `'` at the beginning.
The csv could be used for both excel and something else. But since it's an option...

greg0ire:
By the way, are the headers considered safe?

timwentzell:
> By the way, are the headers considered safe?

Well the headers are normally dependent on the Entity fields and translation labels - so IMO they would be safe.

greg0ire:
Then it's slightly more complex than what I proposed: you would have to take into account `showHeaders` when implementing `SafeCellWriterDecorator::write`

VincentLanglet:
What is needed to merge this @greg0ire ?

greg0ire:
@VincentLanglet IMO src/Writer/CsvWriter.php should not be touched, and this should be implemented with service decoration instead, as show in the snippet above, only it should also take into account `showHeaders` to decide whether to act on the first row or not.

VincentLanglet:
@timwentzell Are you still interested in this PR ?

src/Writer/CsvWriter.php
```diff
@@ -137,6 +144,17 @@ public function write(array $data): void
             EXCEPTION);
         }
 
+        // prevent csv injection
+        if (true === $this->safeCells) {
+            foreach ($data as $key => $value) {
+                $data[$key] = preg_replace(
+                    ['/^=/', '/^\+/', '/^-/', '/^@/'],
```
jderusse, line 151: missing `\r` and `\t` si Symfony's CsvEncoder for example

PR #475, opened 2021-04-08
Fix phpdoc of AbstractPropertySourceIterator::getValue

VincentLanglet:
## Subject

I am targeting this branch, because patch.

For instance, Excel is handling differently string and float values ; so we should not restrict the value to be string.

## Changelog

```markdown
### Fixed
- Allow `AbstractPropertySourceIterator::getValue()` to return `bool|int|float` value.
```

### DIFF
--- a/src/Source/AbstractPropertySourceIterator.php
+++ b/src/Source/AbstractPropertySourceIterator.php
@@ -155,7 +155,7 @@ abstract class AbstractPropertySourceIterator implements SourceIteratorInterface
     /**
      * @param mixed $value
      *
-     * @return string|null
+     * @return bool|int|float|string|null
      */
     protected function getValue($value)
     {


jordisala1991 (approved):

src/Source/AbstractPropertySourceIterator.php
```diff
@@ -155,7 +155,7 @@ protected function getCurrentData($current): array
     /**
      * @param mixed $value
      *
-     * @return string|null
+     * @return bool|int|float|string|null
```
phansys, line 158: In the `default` case, we are returning the same value received in the argument 1. Should we use the same types in its `@param` docblock?

phansys (approved):

src/Source/AbstractPropertySourceIterator.php
VincentLanglet: They are included in mixed. And I think we're supporting other types.
For instance, object are casted to string.

PR #485, opened 2021-06-21
Improve ORM iterator performances

EmmanuelVella:
## Add a batch size for entity manager clear

Clearing the entity manager at each result leads to very poor performances. Using a batch size of 100 seems reasonnable to keep good performances and memory under control.

I am targeting this branch, because it's a BC performance fix.

## Changelog
```markdown
- Clear entity manager every 100 results to improve ORM iterator performances
```

### DIFF
--- a/src/Source/DoctrineORMQuerySourceIterator.php
+++ b/src/Source/DoctrineORMQuerySourceIterator.php
@@ -26,15 +26,22 @@ class DoctrineORMQuerySourceIterator extends AbstractPropertySourceIterator impl
     protected $query;
 
     /**
+     * @var int
+     */
+    private $batchSize;
+
+    /**
      * @param array<string> $fields Fields to export
      */
-    public function __construct(Query $query, array $fields, string $dateTimeFormat = 'r')
+    public function __construct(Query $query, array $fields, string $dateTimeFormat = 'r', int $batchSize = 100)
     {
         $this->query = clone $query;
         $this->query->setParameters($query->getParameters());
         foreach ($query->getHints() as $name => $value) {
             $this->query->setHint($name, $value);
         }
+
+        $this->batchSize = $batchSize;
 
         parent::__construct($fields, $dateTimeFormat);
     }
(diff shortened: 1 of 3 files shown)


src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -25,6 +25,8 @@ class DoctrineORMQuerySourceIterator extends AbstractPropertySourceIterator impl
      */
     protected $query;
 
+    private $batchSize;
```
core23, line 28: Can you please add typehints (with phpdoc)

core23 (changes requested):

src/Source/DoctrineORMQuerySourceIterator.php
```diff
@@ -36,6 +41,8 @@ public function __construct(Query $query, array $fields, string $dateTimeFormat
             $this->query->setHint($name, $value);
         }
 
+        $this->batchSize = 100;
```
VincentLanglet, line 44: What about adding the value to the constructor ?

EmmanuelVella: Ok, it's done !

EmmanuelVella: Hum do you think I should add it to `AbstractPropertySourceIterator` or not ?

VincentLanglet: No because the ODM does not clear https://github.com/sonata-project/exporter/blob/2.x/src/Source/DoctrineODMQuerySourceIterator.php

EmmanuelVella: Yeah I was actually wondering if you would like to clear the ODM too ? But I won't be able to test the changes.

VincentLanglet (dismissed):

phansys (approved):
It would be nice to have a test covering this feature.

core23 (approved):

EmmanuelVella:
> It would be nice to have a test covering this feature.

Added !

VincentLanglet (approved):

phansys (approved):

OskarStark (approved):

OskarStark:
Thanks @EmmanuelVella 🚀

EmmanuelVella:
Thank you for the merge ! Is it possible to tag a new release ?

PR #488, opened 2021-06-28
docs: symfony example with exporter

soullivaneuh:
## Subject

I am targeting this branch, because the only usage we have is for the standalone library directly using the writer.

However, we have a much more simple way to implement an export HTTP response thanks to the exporter service, but I see no example about it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ composer require sonata-project/exporter
 
 ## Usage
 
+### Standalone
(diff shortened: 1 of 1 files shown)


core23 (approved):

phansys (approved):

phansys:
Thank you @soullivaneuh!

PR #519, opened 2021-09-16
Bump phpstan to level 2

### LINKED ISSUE #459: Replace Query::iterable with Query::toIterable
## Feature Request

[`Query::iterable()` is deprecated (`doctrine/orm`) in favor of `Query::toIterable()`](https://github.com/doctrine/orm/blob/5247c56fce2285ba0b0584cfe9376c68a959d012/UPGRADE.md#deprecated-doctrineormabstractqueryiterator):

https://github.com/sonata-project/exporter/blob/03b219538464e588c978a9a634d568dc8ba77748/src/Source/DoctrineORMQuerySourceIterator.php#L55

The difference is that `toIterable` doesn't return an array, so:

https://github.com/sonata-project/exporter/blob/03b219538464e588c978a9a634d568dc8ba77748/src/Source/DoctrineORMQuerySourceIterator.php#L46

`$current[0]` won't be needed, just `$current`.

Also we're using `IterableResult` which is deprecated (and from `doctrine/orm`), I guess with `\Iterator` will be enough.

https://github.com/sonata-project/exporter/blob/03b219538464e588c978a9a634d568dc8ba77748/src/Source/AbstractPropertySourceIterator.php#L36

VincentLanglet:
## Subject

BC, except for the static analysis.
But a generic array-shape is not supported by phpstan, so this wasn't fully working anymore.
Better to remove the template and then fix the static analysis in SonataAdmin.

Also Closes #459.

## Changelog

```markdown
### Changed
- `SourceIteratorInterface` is not generic anymore

### Fixed
- doctrine/orm deprecation
```

### DIFF
--- a/src/Source/SourceIteratorInterface.php
+++ b/src/Source/SourceIteratorInterface.php
@@ -14,8 +14,7 @@ declare(strict_types=1);
 namespace Sonata\Exporter\Source;
 
 /**
- * @phpstan-template T of array
- * @phpstan-extends \Iterator<T>
+ * @phpstan-extends \Iterator<array<mixed>>
  */
 interface SourceIteratorInterface extends \Iterator
 {
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -2,7 +2,7 @@ includes:
     - phpstan-baseline.neon
 
 parameters:
-    level: 1
+    level: 2
     bootstrapFiles:
         - vendor/bin/.phpunit/phpunit/vendor/autoload.php
     paths:
--- a/src/Source/DoctrineORMQuerySourceIterator.php
+++ b/src/Source/DoctrineORMQuerySourceIterator.php
@@ -50,7 +50,7 @@ class DoctrineORMQuerySourceIterator extends AbstractPropertySourceIterator impl
     {
         $current = $this->iterator->current();
 
-        $data = $this->getCurrentData($current[0]);
+        $data = $this->getCurrentData($current);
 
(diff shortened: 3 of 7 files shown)


core23 (approved):

src/Source/SourceIteratorInterface.php
```diff
@@ -14,8 +14,7 @@
 namespace Sonata\Exporter\Source;
 
 /**
- * @phpstan-template T of array
- * @phpstan-extends \Iterator<T>
+ * @phpstan-extends \Iterator<array<mixed>>
```
phansys, line 17: Couldn't we use the template for the array elements?
```
@phpstan-template T
@phpstan-extends \Iterator<array<array-key, T>>
```

VincentLanglet: If you return `array{0: int, 1: string}`, you'll have to use `T = int|string`.

The day support for `array-shape` in template will be supported, we'll have to make a BC-break change from `T = int|string` to `T = array<int|string>` (or `T = array{0: int, 1: string}`) ; so I'm not sure about the benefit.

The other option was to write
```
@phpstan-template T
@phpstan-extends \Iterator<T>
```
and not bound T to array.

But using `T = array<>` or `T = array{}` is not well supported, so it will still be kinda useless.

IMHO silencing this generic could be a simple solution for now. I'm pretty sure this generic will (almost) never be used ; for instance in SonataAdmin, we used `SourceIteratorInterface<array<mixed>>`.

phansys (approved):

PR #560, opened 2021-12-10
Add XLSX writer

### LINKED ISSUE #493: XLS Export - lack of support on Numbers app on OSX due to incorrect file format
## Feature Request

When I download a list in XLS format, the file uses formatting with <td> and other HTML tags. When I try to open it with Numbers app on OSX (macOS), it gives me wrong content.
sonata xls Numbers
Ms Excel can handle the format with <td> but it shouts about incorrect format anyway.

The request is to add a better support for XLS files. @VincentLanglet has helped me a lot on Slack and has suggested to use https://packagist.org/packages/box/spout package, which supports XLSX format. Files may have .xls extension and Numbers still can open the format created by box/spout. Ms Excel can do that on Windows and OSX as well.

The Writer file that I use is following (it may be replaced in the repository, just change getFormat() method).

```php
<?php

namespace Sonata\Exporter\Writer;

use Box\Spout\Writer\Common\Creator\WriterEntityFactory;
use Sonata\Exporter\Writer\TypedWriterInterface;

class XlsWriter implements TypedWriterInterface
{
    /**
     * @var string
     */
    protected $filename;

    /**
     * @var bool
     */
    protected $showHeaders;

    /**
     * @var int
     */
    protected $position;

    /**
     * @var \Box\Spout\Writer\XLSX\Writer
     */
    protected $writer;

    /**
     * @throws \RuntimeException
     */
    public function __construct(string $filename, bool $showHeaders = true)
    {
        $this->filename = $filename;
        $this->showHeaders = $showHeaders;
        $this->position = 0;

        if (is_file($filename)) {
            throw new \RuntimeException(sprintf('The file %s already exist', $filename));
        }
    }

    /**
     * {@inheritdoc}
     */
    final public function getDefaultMimeType(): string
    {
        return 'application/vnd.ms-excel';
    }

    /**
     * {@inheritdoc}
     */
    final public function getFormat(): string
    {
        return 'xls-real';
    }

    /**
     * {@inheritdoc}
     */
    public function open()
    {
        $this->writer = WriterEntityFactory::createXLSXWriter();
        $this->writer->openToFile($this->filename);
    }

    /**
     * {@inheritdoc}
     */
    public function close()
    {
        $this->writer->close();
    }

    /**
     * {@inheritdoc}
     */
    public function write(array $data)
    {
        $this->init($data);

        $row = WriterEntityFactory::createRowFromArray(array_values($data));
        $this->writer->addRow($row);

        ++$this->position;
    }

    protected function init(array $data): void
    {
        if ($this->position > 0) {
            return;
        }

        if ($this->showHeaders) {
            $row = WriterEntityFactory::createRowFromArray(array_keys($data));
            $this->writer->addRow($row);
            ++$this->position;
        }
    }
}
```

willemverspyck:
## Subject

I am targeting this branch, because it is BC.

Closes #493

## Changelog

```markdown
### Added
- Added class `XlsxWriter` that uses "phpoffice/phpspreadsheet" as suggested package. 
- Default the XLSX export is not enabled. You can enable it by adding "xlsx" to "default_writers" in configuration.
- Added tests for the `XlsxWriter` class.
```


tests/Writer/XlsxWriterTest.php
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Sonata\Exporter\Tests\Writer;
+
+use PhpOffice\PhpSpreadsheet\IOFactory;
+use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use Sonata\Exporter\Test\AbstractTypedWriterTestCase;
+use Sonata\Exporter\Writer\TypedWriterInterface;
+use Sonata\Exporter\Writer\XlsxWriter;
+
+class XlsxWriterTest extends AbstractTypedWriterTestCase
```
core23:
```suggestion
final class XlsxWriterTest extends AbstractTypedWriterTestCase
```

composer.json
```diff
@@ -30,6 +31,7 @@
         "doctrine/orm": "^2.9",
         "matthiasnoback/symfony-config-test": "^4.2",
         "matthiasnoback/symfony-dependency-injection-test": "^4.0",
+        "phpoffice/phpspreadsheet": "^1.0",
```
core23, line 34: Can you add the lowest version to the `conflict` section

core23 (changes requested):

composer.json
willemverspyck: @core23: The lowest version is 1.0 and I tested it with version 1.0.0. Is this still needed then?

VincentLanglet: To me, no

src/Writer/XlsxWriter.php
```diff
@@ -0,0 +1,224 @@
...
+     */
+    public function __construct(string $filename, bool $showHeaders = true, bool $showFilters = true)
+    {
+        $this->filename = $filename;
+        $this->showHeaders = $showHeaders;
+        $this->showFilters = $showFilters;
+        $this->position = 1;
+
+        if (!class_exists(Spreadsheet::class)) {
+            throw new RuntimeException('You need the "phpoffice/spreadsheet" component in order to use the XLSX export.');
```
VincentLanglet: To me a runtime exception is expected to be cought. So i would consider more this as a logicException

willemverspyck: You're right @VincentLanglet. Just changed it.

core23 (dismissed):

VincentLanglet:
Test failing

Also, does the doc need to be updated ?

VincentLanglet:
I didn't notice the `default_writers` needed to be updated by the user.

Can we detect the `phpoffice/phpspreadsheet` library and automatically add `xlsx` to the default writers ?

If not, we should at least talk about the `default_writers` option in the doc in the xlsx section.

willemverspyck:
Hi @VincentLanglet. I'm checking now if the class exists and add it to the default_writers. So no user action is required anymore, except installing the package.

VincentLanglet (dismissed):

src/Writer/XlsxWriter.php
```diff
@@ -0,0 +1,222 @@
...
+        $this->showHeaders = $showHeaders;
+        $this->showFilters = $showFilters;
+        $this->position = 1;
+
+        if (!class_exists(Spreadsheet::class)) {
+            throw new \LogicException('You need the "phpoffice/spreadsheet" component in order to use the XLSX export.');
+        }
+
+        if (is_file($filename)) {
+            throw new \RuntimeException(sprintf('The file "%s" already exist', $filename));
```
phansys:
```suggestion
            throw new \RuntimeException(sprintf('The file "%s" already exists.', $filename));
```

src/Writer/XlsxWriter.php
```diff
@@ -0,0 +1,222 @@
...
+     */
+    public function __construct(string $filename, bool $showHeaders = true, bool $showFilters = true)
+    {
+        $this->filename = $filename;
+        $this->showHeaders = $showHeaders;
+        $this->showFilters = $showFilters;
+        $this->position = 1;
+
+        if (!class_exists(Spreadsheet::class)) {
+            throw new \LogicException('You need the "phpoffice/spreadsheet" component in order to use the XLSX export.');
```
phansys:
```suggestion
            throw new \LogicException('You need the "phpoffice/spreadsheet" package in order to use the XLSX export.');
```

src/Writer/XlsxWriter.php
```diff
@@ -0,0 +1,222 @@
...
+     * @throws \RuntimeException
+     */
+    public function __construct(string $filename, bool $showHeaders = true, bool $showFilters = true)
+    {
+        $this->filename = $filename;
+        $this->showHeaders = $showHeaders;
+        $this->showFilters = $showFilters;
+        $this->position = 1;
+
+        if (!class_exists(Spreadsheet::class)) {
```
phansys, line 65: For consistency, I think we should make all the checks previous to any assignment.

tests/Writer/XlsxWriterTest.php
```diff
@@ -0,0 +1,98 @@
...
+namespace Sonata\Exporter\Tests\Writer;
+
+use PhpOffice\PhpSpreadsheet\IOFactory;
+use Sonata\Exporter\Test\AbstractTypedWriterTestCase;
+use Sonata\Exporter\Writer\TypedWriterInterface;
+use Sonata\Exporter\Writer\XlsxWriter;
+
+final class XlsxWriterTest extends AbstractTypedWriterTestCase
+{
+    protected $filename;
```
phansys: If this class is final, why these properties need to be protected?

phansys (changes requested):

tests/Writer/XlsxWriterTest.php
VincentLanglet: Certainly copied pasted from another test but indeed it can be private

src/Bridge/Symfony/DependencyInjection/SonataExporterExtension.php
```diff
@@ -39,7 +39,7 @@ public function load(array $configs, ContainerBuilder $container): void
 
     private function configureExporter(ContainerBuilder $container, array $config): void
     {
-        foreach (['csv', 'json', 'xls', 'xml'] as $format) {
+        foreach (['csv', 'json', 'xls', 'xlsx', 'xml'] as $format) {
```
franmomu, line 42: ~Not related to this PR, but IMO we should add the tags in the definitions instead of updating two places every time we add a writer, I'll make a PR.~

Nevermind, not applicable right now.

src/Bridge/Symfony/Resources/config/services.xml
```diff
@@ -16,6 +16,11 @@
             <argument>%sonata.exporter.writer.xls.filename%</argument>
             <argument>%sonata.exporter.writer.xls.show_headers%</argument>
         </service>
+        <service id="sonata.exporter.writer.xlsx" class="Sonata\Exporter\Writer\XlsxWriter" public="false">
```
franmomu: IMO the service could also be loaded conditionally, otherwise I'm not sure if it will fail if the package is not installed (it's throwing an exception in the constructor)

src/Writer/XlsxWriter.php
```diff
@@ -0,0 +1,222 @@
...
+
+            $this->worksheet->setCellValueExplicitByColumnAndRow($column, $this->position, $dataValue, $dataType);
+
+            ++$column;
+        }
+
+        ++$this->position;
+    }
+
+    private function addHeaders(array $data): void
```
franmomu, line 143:
```suggestion
    /**
     * @param array<string, mixed> $data
     */
    private function addHeaders(array $data): void
```

src/Writer/XlsxWriter.php
```diff
@@ -0,0 +1,222 @@
...
+            $this->worksheet->setCellValueExplicitByColumnAndRow($column, $this->position, $value, DataType::TYPE_STRING);
+
+            ++$column;
+        }
+    }
+
+    /**
+     * Get the type of the Spreadsheet cell.
+     */
+    private function getDataType($value): string
```
franmomu, line 159: can we add `mixed` as parameter type in phpdoc? (this applies to other `private` functions)

tests/Writer/XlsxWriterTest.php
willemverspyck: Yes, copied from the other tests. Made them private now.

src/Writer/XlsxWriter.php
willemverspyck: Done

src/Bridge/Symfony/Resources/config/services.xml
willemverspyck: Do you mean by checking the if the Spreadsheet class exists in the services.xml (now services.php) and then add the service?

tests/Writer/XlsxWriterTest.php
```diff
@@ -0,0 +1,98 @@
...
+namespace Sonata\Exporter\Tests\Writer;
+
+use PhpOffice\PhpSpreadsheet\IOFactory;
+use Sonata\Exporter\Test\AbstractTypedWriterTestCase;
+use Sonata\Exporter\Writer\TypedWriterInterface;
+use Sonata\Exporter\Writer\XlsxWriter;
+
+final class XlsxWriterTest extends AbstractTypedWriterTestCase
+{
+    private $filename;
```
phansys, line 26:
```suggestion
    /**
     * @var string
     */
    private $filename;
```

tests/Writer/XlsxWriterTest.php
```diff
@@ -0,0 +1,98 @@
...
+    private $filenameCsv;
```
phansys, line 31:
```suggestion
    /**
     * @var string
     */
    private $filenameCsv;
```

tests/Writer/XlsxWriterTest.php
```diff
@@ -0,0 +1,98 @@
...
+        $excelWriter->save($this->filenameCsv);
+
+        $expected = sprintf('"firtname","surname","year"%s"john ""2","doe","1"%s', \PHP_EOL, \PHP_EOL);
+
+        static::assertSame($expected, file_get_contents($this->filenameCsv));
+    }
+
+    protected function getWriter(): TypedWriterInterface
+    {
+        return new XlsxWriter('/tmp/whatever.xlsx', false);
```
phansys:
```suggestion
        return new XlsxWriter(sprintf('%s/whatever.xlsx', sys_get_temp_dir()), false);
```

src/Writer/XlsxWriter.php
```diff
@@ -0,0 +1,233 @@
...
+        if (!class_exists(Spreadsheet::class)) {
+            throw new \LogicException('You need the "phpoffice/spreadsheet" package in order to use the XLSX export.');
+        }
+
+        $this->filename = $filename;
+        $this->showHeaders = $showHeaders;
+        $this->showFilters = $showFilters;
+        $this->position = 1;
+
+        if (is_file($filename)) {
```
phansys: Could you please move this check before the assignments?

src/Writer/XlsxWriter.php
```diff
@@ -0,0 +1,233 @@
...
+            throw new \RuntimeException(sprintf('The file "%s" already exists.', $filename));
```
phansys:
```suggestion
            throw new \InvalidArgumentException(sprintf('The file "%s" already exists.', $filename));
```

src/Writer/XlsxWriter.php
```diff
@@ -0,0 +1,233 @@
...
+    private $worksheet;
+
+    /**
+     * @var int
+     */
+    private $position;
+
+    /**
+     * @throws \LogicException
+     * @throws \RuntimeException
```
phansys:
```suggestion
     * @throws \InvalidArgumentException
```

phansys (changes requested):

src/Bridge/Symfony/Resources/config/services.xml
franmomu: Yeah, before I was thinking about doing it in the `SonataExporterExtension.php`, but now we can do it in `services.php`.

src/Writer/XlsxWriter.php
willemverspyck: Thanks @phansys. All changes are fixed 👍

src/Bridge/Symfony/Resources/config/services.xml
willemverspyck: Perfect. Just fixed this 👍

phansys (dismissed):

VincentLanglet (dismissed):

src/Bridge/Symfony/Resources/config/services.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
franmomu: Please remove this file

franmomu (changes requested):

VincentLanglet (approved):

franmomu (approved):

phansys (approved):

phansys:
Thanks @willemverspyck!

PR #550, opened 2021-11-12
Fix compatibility with Doctrine DBAL 2.13

jordisala1991:
## Subject

I am targeting this branch, because this is BC.

Closes #549 .

## Changelog

```markdown
### Fixed
- Fixed compatibility with Doctrine DBAL ^2.13.
```

PR #540, opened 2021-10-24
Add compatibility with Doctrine DBAL 3

jordisala1991:
## Subject

I am targeting this branch, because this is BC.

## Changelog

```markdown
### Added
- Added compatibility with Doctrine DBAL 3.
```

PR #562, opened 2021-12-19
Bump static analysis

VincentLanglet:
## Changelog

```markdown
### Fixed
- DBAL v3 compatibility
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,5 +98,5 @@ phpstan:
 .PHONY: phpstan
 
 psalm:
-	vendor/bin/psalm
+	vendor/bin/psalm --php-version=8.1
 .PHONY: psalm
--- a/src/Source/DoctrineDBALConnectionSourceIterator.php
+++ b/src/Source/DoctrineDBALConnectionSourceIterator.php
@@ -61,6 +61,7 @@ final class DoctrineDBALConnectionSourceIterator implements SourceIteratorInterf
     /**
      * @return array<string, mixed>
      */
+    #[\ReturnTypeWillChange]
     public function current()
     {
         return $this->current;
@@ -73,8 +74,9 @@ final class DoctrineDBALConnectionSourceIterator implements SourceIteratorInterf
     }
 
     /**
-     * @return mixed
+     * @return int
      */
+    #[\ReturnTypeWillChange]
     public function key()
     {
         return $this->position;
@@ -85,6 +87,9 @@ final class DoctrineDBALConnectionSourceIterator implements SourceIteratorInterf
         return \is_array($this->current);
     }
 
+    /**
+     * @psalm-suppress InvalidPropertyAssignmentValue
+     */
     public function rewind(): void
     {
         $statement = $this->connection->prepare($this->query);
--- a/src/Source/DoctrineDBALConnectionSourceIterator.php
+++ b/src/Source/DoctrineDBALConnectionSourceIterator.php
@@ -89,11 +89,14 @@ final class DoctrineDBALConnectionSourceIterator implements SourceIteratorInterf
     {
         $statement = $this->connection->prepare($this->query);
 
-        // TODO: Keep only the else part when dropping support for Doctrine DBAL < 3.1
-        if (method_exists($statement, 'executeQuery')) {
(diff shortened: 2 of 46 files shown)


Makefile
```diff
@@ -98,5 +98,5 @@ phpstan:
 .PHONY: phpstan
 
 psalm:
-	vendor/bin/psalm
+	vendor/bin/psalm --php-version=8.1
```
VincentLanglet, line 101: https://github.com/sonata-project/dev-kit/pull/1828

src/Source/DoctrineDBALConnectionSourceIterator.php
```diff
@@ -80,13 +80,11 @@ public function rewind(): void
...
 
-        // TODO: Keep only the if part when dropping support for Doctrine DBAL < 3.1
+        // TODO: Keep only the else part when dropping support for Doctrine DBAL < 3.1
         if (method_exists($statement, 'executeQuery')) {
             $this->result = $statement->executeQuery($this->parameters);
         } else {
-            $statement->execute($this->parameters);
-
-            $this->result = $statement;
+            $this->result = $statement->execute($this->parameters);
```
VincentLanglet: Can you take a look @jordisala1991, I feel like the DBAL3 code was wrong in your fix https://github.com/sonata-project/exporter/pull/550/files

jordisala1991: I think the older version returns void or bool and you have to use the $statement (static analysis wont help here since they only run with highest deps)

src/Source/DoctrineDBALConnectionSourceIterator.php
```diff
@@ -80,13 +80,11 @@ public function rewind(): void
     {
         $statement = $this->connection->prepare($this->query);
 
-        // TODO: Keep only the if part when dropping support for Doctrine DBAL < 3.1
+        // TODO: Keep only the else part when dropping support for Doctrine DBAL < 3.1
```
jordisala1991: The if part is the good one

VincentLanglet: I need to check tomorrox but I thought I had dbal 3.2 and no executeQuery method

src/Source/DoctrineDBALConnectionSourceIterator.php
VincentLanglet: When I look at https://github.com/sonata-project/exporter/pull/540/files#diff-38640adca7b4122bbd27d6173df4371be37ef41c1e885c744844e67755b0d35dR82
To me, the older version was returning void and in 3.0 it returns a result.

Then https://github.com/sonata-project/exporter/pull/550/files was made to fix the 2.13 compatibility, so you added the if part with the executeQuery for the version 2. In the futur we need to keep the else part.

src/Source/DoctrineDBALConnectionSourceIterator.php
VincentLanglet: https://github.com/sonata-project/exporter/pull/562/files#r772146858

I think it's the else part we need to keep.
I have dbal 3.2 locally and no executeQuery method

src/Source/DoctrineDBALConnectionSourceIterator.php
jordisala1991: Can you try to add a test that covers those lines?

VincentLanglet: It require a `Doctrine\DBAL\Driver\Connection` and I dunno how to have one.

VincentLanglet:
@jordisala1991 @sonata-project/contributors :)

src/Source/DoctrineDBALConnectionSourceIterator.php
jordisala1991: I think it is provided by driver manager? https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html

VincentLanglet: Done ! And I find the right fix

jordisala1991 (approved):

PR #567, opened 2022-01-03
Fix SourceIteratorInterface generic

VincentLanglet:
## Changelog

```markdown
### Fixed
- SourceIteratorInterface unresolvable generics
```

### DIFF
--- a/src/Source/ChainSourceIterator.php
+++ b/src/Source/ChainSourceIterator.php
@@ -13,6 +13,9 @@ declare(strict_types=1);
 
 namespace Sonata\Exporter\Source;
 
+/**
+ * @phpstan-implements SourceIteratorInterface<mixed, mixed>
+ */
 final class ChainSourceIterator implements SourceIteratorInterface
 {
     /**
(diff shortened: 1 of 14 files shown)


src/Source/ChainSourceIterator.php
```diff
@@ -16,10 +16,13 @@
 final class ChainSourceIterator implements SourceIteratorInterface
 {
     /**
-     * @var \ArrayIterator
+     * @var \ArrayIterator<array-key, SourceIteratorInterface>
      */
     private $sources;
 
+    /**
+     * @param array<SourceIteratorInterface> $sources
```
jordisala1991, line 24: Do you know if there is a php fixer rule to avoid having both: Foo[] and array<Foo> in the codebase?

jordisala1991 (approved):

src/Source/ChainSourceIterator.php
VincentLanglet: I would say there is none

PR #583, opened 2022-05-22
Add rector and drop PHP 7.3

VincentLanglet:
## Subject

I am targeting this branch, because {reason}.

Closes #{put_issue_number_here}.

## Changelog

```markdown
### Removed
- Support for PHP 7.3
```


src/Writer/GsaFeedWriter.php
```diff
@@ -20,32 +20,15 @@
  */
 final class GsaFeedWriter implements WriterInterface
 {
-    public const LIMIT_SIZE = 31457280; // 30MB
+    public const LIMIT_SIZE = 31_457_280; private \SplFileInfo $folder;
```
VincentLanglet, line 23: Weird

src/Writer/SitemapWriter.php
```diff
@@ -18,28 +18,16 @@
  */
 final class SitemapWriter implements WriterInterface
 {
-    public const LIMIT_SIZE = 10485760;
+    public const LIMIT_SIZE = 10_485_760;
```
VincentLanglet, line 21: Weird

src/Writer/XlsxWriter.php
```diff
@@ -26,35 +26,17 @@
...
-     * @var Spreadsheet
-     */
-    private $spreadsheet;
+    private ?Spreadsheet $spreadsheet = null;
 
-    /**
-     * @var Worksheet
-     */
-    private $worksheet;
+    private ?Worksheet $worksheet = null;
```
jordisala1991, line 37: Is correct to change both properties to nullable?

VincentLanglet: The property is not initialized in the constructor so far.
Phpstan is complaining:
```
Class Sonata\Exporter\Writer\XlsxWriter has an uninitialized property $spreadsheet. Give it default value or assign it in the constructor.
```
If you call `write()` without `open()` before, you'll end with an access to an uninitialized property.

Later we will need to add some null check and throw some exceptions.

jordisala1991 (approved):

PR #595, opened 2022-05-30
Deprecate passing a DriverConnection

VincentLanglet:
## Subject

Related to the deprecation in dbal https://github.com/doctrine/dbal/pull/4966

It will work this way in 3.x https://github.com/sonata-project/exporter/pull/594

## Changelog

```markdown
### Deprecated
- Passing  an instance of `Doctrine\DBAL\Driver\Connection` to `DoctrineDBALConnectionSourceIterator::__construct()`
```

### DIFF
--- a/src/Source/DoctrineDBALConnectionSourceIterator.php
+++ b/src/Source/DoctrineDBALConnectionSourceIterator.php
@@ -13,12 +13,17 @@ declare(strict_types=1);
 
 namespace Sonata\Exporter\Source;
 
-use Doctrine\DBAL\Driver\Connection;
-use Doctrine\DBAL\Driver\Result;
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\Connection as DriverConnection;
+use Doctrine\DBAL\Driver\Result as DriverResult;
+use Doctrine\DBAL\Result;
 
 final class DoctrineDBALConnectionSourceIterator implements SourceIteratorInterface
 {
-    private Connection $connection;
+    /**
+     * @var Connection|DriverConnection
+     */
+    private $connection;
 
     private string $query;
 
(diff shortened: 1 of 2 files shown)


src/Source/DoctrineDBALConnectionSourceIterator.php
```diff
@@ -35,15 +40,25 @@ final class DoctrineDBALConnectionSourceIterator implements SourceIteratorInterf
...
+    public function __construct($connection, string $query, array $parameters = [])
     {
+        if (!$connection instanceof Connection) {
+            @trigger_error(sprintf(
+                'Passing an instance of %s as argument 1 is deprecated since sonata-project/exporter 2.13'
+                .' and will not work in 3.0. You MUST pass an instance of %s instead',
+                \get_class($connection),
+                Connection::class
+            ), \E_USER_DEPRECATED);
+        }
```
derrabus:
```suggestion
        if ($connection instanceof DriverConnection) {
            @trigger_error(sprintf(
                'Passing an instance of %s as argument 1 is deprecated since sonata-project/exporter 2.13'
                .' and will not work in 3.0. You MUST pass an instance of %s instead',
                \get_class($connection),
                Connection::class
            ), \E_USER_DEPRECATED);
        } elseif (!$connection instanceof Connection) {
            throw new \TypeError(sprintf(
                '%s: Argument 1 is expected to be an instance of %s, got %s.',
                __METHOD__,
                Connection::class,
                get_class($connection)
            ));
        }
```

You can still throw if `$connection` is not `Connection|DriverConnection` as you did before.

VincentLanglet: Indeed. I just need to do differently since in DBAL 2 the connection implements DriverConnection
https://github.com/doctrine/dbal/blob/v2.10.1/lib/Doctrine/DBAL/Connection.php#L37

jordisala1991 (approved):

fezfez:
Hi, can it be release ? Thanks in advance

VincentLanglet:
> Hi, can it be release ? Thanks in advance

On it https://github.com/sonata-project/exporter/pull/597

PR #618, opened 2023-01-02
Added ExporterInterface

pkameisha:
## Subject

I am targeting this branch, because we need to follow dependence inversion principle. CRUDController in sonata-project/SonataAdminBundle shouldn't depend on Exporter class, but on ExporterInterface

## Changelog

```markdown
### Added
- Added ExporterInterface;
- Added implementation ExporterInterface by Exporter.
```

### DIFF
--- a/src/ExporterInterface.php
+++ b/src/ExporterInterface.php
@@ -2,6 +2,15 @@
 
 declare(strict_types=1);
 
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 namespace Sonata\Exporter;
 
 use Symfony\Component\HttpFoundation\StreamedResponse;
--- a/src/ExporterInterface.php
+++ b/src/ExporterInterface.php
@@ -1,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sonata\Exporter;
+
+use Symfony\Component\HttpFoundation\StreamedResponse;
+
+interface ExporterInterface
+{
+    /**
+     * @return string[]
+     */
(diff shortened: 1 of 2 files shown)


src/ExporterInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+declare(strict_types=1);
+
+namespace Sonata\Exporter;
+
+use Symfony\Component\HttpFoundation\StreamedResponse;
+
+interface ExporterInterface
+{
+    public function getResponse(string $format, string $filename, \Iterator $source): StreamedResponse;
```
VincentLanglet, line 25: Here: https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Bridge/Exporter/AdminExporter.php#L24-L54

we're also using `getAvailableFormats`, so I think we should add it to the Interface too.

VincentLanglet (changes requested):

src/ExporterInterface.php
pkameisha: Fixed.  Ok, I see, I will update AdminExporter too.

src/ExporterInterface.php
```diff
@@ -0,0 +1,14 @@
...
+
+declare(strict_types=1);
+
+namespace Sonata\Exporter;
+
+use Symfony\Component\HttpFoundation\StreamedResponse;
+
+interface ExporterInterface
+{
+    public function getAvailableFormats(): array;
```
core23, line 23: Type hints are missing for the array. Maybe `string[]`

VincentLanglet, line 23:
```suggestion
    /**
     * @return string[]
     */
    public function getAvailableFormats(): array;
```
to avoid the PHPStan error

pkameisha: done

VincentLanglet (dismissed):

VincentLanglet:
Failure are unrelated

VincentLanglet:
Thanks, I'll make a release

PR #630, opened 2023-03-19
Update psalm to v5, drop sf 6.0 and 6.1

jordisala1991:
## Subject

I am targeting this branch, because this is pedantic. Symfony deps are dev dependencies, except for symfony config, which is already on conflict section.


src/Source/AbstractXmlSourceIterator.php
```diff
@@ -163,8 +163,13 @@ final protected function parseRow(): void
 
         $this->currentRowEnded = false;
 
-        // read file until row is ended
-        // @phpstan-ignore-next-line: The currentRowEnded value is updated when parsing the data
+        /**
+         * Read file until row is ended
+         * The currentRowEnded value is updated when parsing the data
```
jordisala1991:
```suggestion
         * The currentRowEnded value is updated when parsing the data.
```

jordisala1991:
Do you know how to fix last error or should I supress it? @VincentLanglet it looks like a bug on psalm side.

franmomu:
> Do you know how to fix last error or should I supress it? @VincentLanglet it looks like a bug on psalm side.

I think I would inline the iterator and remove the `testGetIterator` which is already tested in https://github.com/sonata-project/exporter/blob/cea96d6a8619bae185dd0af53566523d9f1656a7/tests/Source/IteratorSourceIteratorTest.php#L35-L38

VincentLanglet:
> Do you know how to fix last error or should I supress it? @VincentLanglet it looks like a bug on psalm side.

It's not a bug on psalm-side. https://github.com/vimeo/psalm/issues/9501 for context.
When I encounter this bug, the solution was to change
```
@var \ArrayIterator<int, array{int}>
```
to
```
@var \Traversable<int, array{int}>
```

But here we require an Iterator... https://github.com/sonata-project/exporter/blob/3f7d3c7356b0b05f587b20a0b196f074a7cdc1af/src/Source/IteratorCallbackSourceIterator.php#L24

VincentLanglet (approved):

jordisala1991:
Using traversable works, but also removing the property iterator as @franmomu suggested, since it is tested on other place. Which solution do you prefer @VincentLanglet ?

VincentLanglet:
> Using traversable works, but also removing the property iterator as @franmomu suggested, since it is tested on other place. Which solution do you prefer @VincentLanglet ?

I'm ok with this one.

